Issue #7, opened 2023-10-30
Filename passed in to download function gets messed up in Chinese


marshallpete:
### Provide a general summary of the issue here

When trying to download a PNG, the provided filename gets transformed by RSC into snake_case. The issue with this is some languages like Chinese, cannot be transformed into snake case as so the resulting filename is sometimes garbage.

### 🤔 Expected Behavior?

Provided filename is used unchanged by RSC.

### 😯 Current Behavior

Provided filename gets transformed into snake_case.

### 💁 Possible Solution

Don't transform into snake_case

### 🖥️ Steps to Reproduce

Try to export a PNG and pass in the name of the file as 'My file'. The output png will be 'my_file.png' when it should just be 'My file.png'

### Version

0.0.1.alpha.4

### What browsers are you seeing the problem on?

Chrome

### What operating system are you using?

Mac

PR #8, opened 2023-10-30
fix: do not use snake case on potentially localized values

marshallpete:
## Description

When exporting a PNG and providing a custom filename for the export, Chinese characters get turned in to non-sense.

## Related Issue

[Filename passed in to download function gets messed up in Chinese](https://github.com/adobe/react-spectrum-charts/issues/7)

## Motivation and Context

This is causing internationalization problems.

## How Has This Been Tested?

`yarn pack-test`
Then installed the tarball in product to verify issue was fixed.

## Types of changes

## Checklist:

### DIFF
--- a/src/Prism.tsx
+++ b/src/Prism.tsx
@@ -135,11 +135,6 @@ export const Prism = forwardRef<PrismHandle, PrismProps>(
 			UNSAFE_vegaSpec,
 		});
 
-		// if the spec has changed, reset the chartView
-		useEffect(() => {
-			chartView.current = undefined;
-		}, [spec]);
-
 		const { controlledHoverSignal, selectedIdSignalName, selectedSeriesSignalName } = useSpecProps(spec);
 		const prismConfig = useMemo(() => getPrismConfig(config, colorScheme), [config, colorScheme]);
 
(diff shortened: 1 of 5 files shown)


src/Prism.tsx
```diff
@@ -135,11 +135,6 @@ export const Prism = forwardRef<PrismHandle, PrismProps>(
 			UNSAFE_vegaSpec,
 		});
 
-		// if the spec has changed, reset the chartView
-		useEffect(() => {
-			chartView.current = undefined;
-		}, [spec]);
-
```
marshallpete, line 142: This isn't needed and was causing issues where the image couldn't be downloaded after a rerender

src/hooks/usePrismImperativeHandle.tsx
```diff
@@ -46,7 +45,7 @@ export default function usePrismImperativeHandle(
 		download(customFileName?: string) {
 			return new Promise<string>((resolve, reject) => {
 				if (chartView.current) {
-					const filename = `${toSnakeCase(customFileName || title || 'prism_export')}.png`;
+					const filename = `${customFileName || title || 'chart_export'}.png`;
```
marshallpete, line 48: We don't want to use snake_case and instead should just honor the filename provided.

evargast (approved):

PR #17, opened 2023-11-07
Switch to vega embed

marshallpete:
## Description

Currently the library uses a tool called react-vega to render the vega charts. The issue with react-vega is we get some weird behavior like resize flicker and a lot of console.errors and console.warns.

Going one level lower to vega-embed and controlling the lifecycle ourselves solves these issues.

## How Has This Been Tested?

All tests are passing which confirms that the new vega renderer was swapped out successfully.

## Types of changes

## Checklist:

### DIFF
--- a/src/Chart.tsx
+++ b/src/Chart.tsx
@@ -141,14 +141,10 @@ export const Chart = forwardRef<ChartHandle, ChartProps>(
 			tooltipElement.hidden = popoverIsOpen;
 
 			// if the popover is closed, reset the selected data
-			if (!popoverIsOpen && chartView.current) {
+			if (!popoverIsOpen) {
 				selectedData.current = null;
-
-				setTimeout(() => {
-					chartView.current?.runAsync();
-				}, 1000);
 			}
-		}, [popoverIsOpen, chartView]);
+		}, [popoverIsOpen]);
 
 		useChartImperativeHandle(forwardedRef, { chartView, title });
 
--- a/src/Chart.tsx
+++ b/src/Chart.tsx
@@ -9,11 +9,11 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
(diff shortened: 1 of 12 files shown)


src/Chart.tsx
```diff
@@ -138,36 +134,26 @@ export const Chart = forwardRef<ChartHandle, ChartProps>(
...
 			tooltipElement.hidden = popoverIsOpen;
 
 			// if the popover is closed, reset the selected data
-			if (!popoverIsOpen) {
+			if (!popoverIsOpen && chartView.current) {
 				selectedData.current = null;
+
+				setTimeout(() => {
+					chartView.current?.runAsync();
+				}, 1000);
```
ToyWalrus, line 149: Is this `1000` just an arbitrary number? Is it worth defining a constant somewhere if we expect to do this (timeouts) in more than one place?

src/Chart.tsx
```diff
@@ -138,36 +134,26 @@ export const Chart = forwardRef<ChartHandle, ChartProps>(
...
 			}
-		}, [popoverIsOpen]);
+		}, [popoverIsOpen, chartView]);
```
ToyWalrus, line 151: `chartView` is a `ref` object, so it will never trigger this `useEffect()`. If you want the `useEffect()` to run when `chartView.current` changes you have a few options:
1. Make `chartView` a stateful variable (probably not desired)
2. Remove `chartView` from the dependency list

There's also a third option which is to keep it in (as `chartView.current`), but be aware that updates to `chartView.current` _won't trigger a rerender_. But if `chartView.current` is different from one render to the next, it will trigger this `useEffect()`.

src/Chart.tsx
marshallpete: I did not mean to leave this in there haha

src/Chart.tsx
marshallpete: This is old and I don't need this in there anymore.

ToyWalrus (approved):

PR #23, opened 2023-11-09
Fix vega dependencies

marshallpete:
## Description

vega and vega-lite were listed as dependencies. The issue with this is it is likely that users of this library will have multiple instances installed of vega/vega-lite and all the associated packages. This is not only bad for size of installed packages, it can cause RSC to fail to render because some of the vega packages are persistent instances. For example, vega-functions us used to register custom expression functions. If the functions is registered to one instance but vega tries to digest the functions from the other instance, the registered function won't exist.

## Motivation and Context

see description

## How Has This Been Tested?

Ran `yarn pack-test` and installed into APA where the multiple install issue was reported from. Confirmed that it is fixed.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,8 @@
 		"ts-loader": "^9.4.2",
 		"tsconfig-paths-webpack-plugin": "^4.0.0",
 		"typescript": "^4.9.3",
+		"vega": "5.26.0",
+		"vega-lite": "^5.16.2",
 		"webpack": "^5.75.0",
 		"webpack-cli": "^5.0.1",
 		"webpack-dev-server": "^4.11.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -103,14 +105,14 @@
 		"react": "17.0.2 - latest",
 		"react-dom": "17.0.2 - latest",
 		"uuid": "9.0.0 - latest",
-		"vega": "5.22.1 - latest",
-		"vega-embed": "6.21.0 - latest",
-		"vega-lite": "5.6.0 - latest"
+		"vega-embed": "6.23.0 - latest"
```
evargast, line 108: Should we add `vega-embed` to the peerDependencies to use that module if already installed?

marshallpete: It shouldn't already be installed by the user because most people shouldn't already be doing vega charts on their own.
I'm kinda following how `vega/react-vega` does it https://github.com/vega/react-vega/blob/master/packages/react-vega/package.json

tanugarg04 (approved):

PR #31, opened 2023-11-14
docs: remove placeholder description and set story height to auto

marshallpete:
Just cleaning up the storybook experience


Bryant-Curtis (approved):
Love it! -216 lines of code!!

PR #59, opened 2024-01-05
fix: enforce minimum vega-tooltip for custom tooltips

rmillio:
This PR only adds a dependency in package.json which is a requirement for RSC but wasn't listed. 

## Description

It all started with a bug in CJA where custom tooltip wouldn't work after upgrading to the latest alpha (alpha-12). After a bit of digging I found out:
* tag v0.0.1-alpha.10 is where the tooltip stops working for me, alpha.9 is ok
* if I comment (in alpha.10) src/Chart.tsx lines 287 through 305, the issue goes away

## Solution
List the package `vega-tooltip` as a dependency since it is a requirement for custom tooltips. The custom tooltip support was added in `vega-tooltip v0.26.0`. For the CJA bug, @quarry/react-vega (a dep of CJA) has vega-tooltip@0.19.0 which explains why the custom tooltip wouldn't work. The solution was to add `vega-tooltip` to resolutions.

## Motivation and Context
Adding vega-tooltip as a dependency will enforce a minimum version which is required for custom tooltips. It will help other consumers avoid dependency mismatch and make RSC more robust.

## How Has This Been Tested?

There are no code changes, this PR only updates `package.json`.

Before:
image

After:
image

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,8 @@
 		"@adobe/react-spectrum": ">= 3.23.0",
 		"immer": ">= 9.0.0",
 		"uuid": ">= 9.0.0",
-		"vega-embed": ">= 6.23.0"
+		"vega-embed": ">= 6.23.0",
+		"vega-tooltip": ">= 0.26.0"
 	},
 	"peerDependencies": {
 		"@adobe/react-spectrum": ">=3.23.0",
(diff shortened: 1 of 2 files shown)


marshallpete:
GET_BUILD

marshallpete:
GET_BUILD

marshallpete:
GET_BUILD

PR #114, opened 2024-02-07
Nathaniel judd/byu team/big number

Judd2000:
Method property for BigNumber

## Description

Added method attribute to BigNumberProps, calculate BigNumber's display value based on the method provided, tests for this as well as fixed tests for BigNumber's Chart.tsx implementation.

## Motivation and Context

Added implementation to the Big-Number component

## How Has This Been Tested?

Jest tests written in BigNumber.test.tsx that pass.

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [X ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [X ] My code follows the code style of this project.
- [X ] My change requires a change to the documentation.
- [X ] I have updated the documentation accordingly.
- [X ] I have read the **CONTRIBUTING** document.
- [ X] I have added tests to cover my changes.
- [ X] All new and existing tests passed.

### DIFF
--- a/src/components/BigNumber/BigNumber.tsx
+++ b/src/components/BigNumber/BigNumber.tsx
@@ -38,10 +38,25 @@ export function BigNumber({
 		chartView: useRef<VegaView>(),
 		chartWidth: 600,
 	},
+	method
 }: BigNumberProps) {
 	const { chartWidth, height, locale, data, ...rscChartRemain } = rscChartProps;
 
-	const bigNumberValue = data && data.length > 0 ? data[data.length - 1][dataKey] : undefined;
+	if (!method) {
+		method = 'last';
+	}
+
+	let bigNumberValue: number | undefined;
+	if (method === 'last') {
+		bigNumberValue = data && data.length > 0 ? data[data.length - 1][dataKey] : undefined;
+	} else if (method === 'sum' || method === 'avg') {
+		bigNumberValue = data && data.length > 0 ? data.reduce((sum, cur) => {
+			return sum + cur[dataKey];
+		}, 0) : undefined;
+		if (method === 'avg') {
+			bigNumberValue = bigNumberValue ? bigNumberValue / data.length : bigNumberValue;
+		}
+	}
 
 	const numberLocale = getLocale(locale).number;
 	const type = numberType ?? 'linear';
@@ -105,7 +120,7 @@ export function BigNumber({
 							locale={locale}
 							{...rscChartRemain}
 						>
-							{lineElements.map((le) => cloneElement(le as LineElement, { isSparkline: true }))}
+							{lineElements.map((le) => cloneElement(le as LineElement, { isSparkline: true, isMethodLast: method === 'last' }))}
 						</RscChart>
 					</View>
 				)}
--- a/src/components/BigNumber/BigNumber.tsx
+++ b/src/components/BigNumber/BigNumber.tsx
@@ -38,13 +38,13 @@ export function BigNumber({
 		chartView: useRef<VegaView>(),
 		chartWidth: 600,
 	},
-	method = 'last'
+	method
 }: BigNumberProps) {
 	const { chartWidth, height, locale, data, ...rscChartRemain } = rscChartProps;
 
 	// based on Chart.tsx checks, data will always be defined and have a length greater than 0.
 	let bigNumberValue: number;
-	if (method === 'last') {
+	if (!method || method === 'last') {
 		bigNumberValue = data[data.length - 1][dataKey];
 	} else {
 		// this must be either 'sum' or 'avg'
--- a/src/components/BigNumber/BigNumber.tsx
+++ b/src/components/BigNumber/BigNumber.tsx
@@ -38,21 +38,19 @@ export function BigNumber({
 		chartView: useRef<VegaView>(),
 		chartWidth: 600,
 	},
-	method
+	method = 'last'
 }: BigNumberProps) {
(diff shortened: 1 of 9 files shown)


src/components/BigNumber/BigNumber.tsx
```diff
@@ -40,10 +38,25 @@ export function BigNumber({
...
+	method
 }: BigNumberProps) {
 	const { chartWidth, height, locale, data, ...rscChartRemain } = rscChartProps;
 
-	const bigNumberValue = data && data.length > 0 ? data[data.length - 1][dataKey] : undefined;
+	if (!method) {
+		method = 'last';
+	}
+
+	let bigNumberValue: number | undefined;
```
jacob-hart, line 49: Could we try refactoring this to make bigNumberValue never undefined?  I'm concerned that having to check for undefined later on makes the code more complex.  It might also make it easier to pass the sonarcloud smells test.  If its too difficult don't worry though

src/stories/components/BigNumber/BigNumber.story.tsx
```diff
@@ -126,45 +179,49 @@ SparklineAndIconVertical.args = {
...
+const CompactFormat = bindWithProps(CompactFormatStory);
 CompactFormat.args = {
 	children: undefined,
-	dataKey: 'x',
+	dataKey: 'value',
 	orientation: 'horizontal',
 	label: 'Requests',
 };
 
-// const NullData = bindWithProps(BigNumberStory);
```
jacob-hart, line 155: Could this case still exist?  If type safety helps us make data never null, then it's probably safe

src/stories/components/BigNumber/BigNumber.test.tsx
```diff
@@ -102,4 +90,68 @@ describe('BigNumber', () => {
 			expect(formattedValue).toBeInTheDocument();
 		});
 	});
+
+	describe('Chart with BigNumber children', () => {
```
jacob-hart, line 94: great job with these cases

jacob-hart (commented):
I left a few comments for review!

Also it would be good to have all the quality gates pass.  That way this commit could potentially go directly to main

src/components/BigNumber/BigNumber.tsx
gideontonkinson: Agreed, on line 41 you can do `method = 'last'` and that will set it to that as a default should it be undefined and you don't need to do a check for it, hence why orientation and rscCharProps both also have defaults so I don't have to an undefined check

src/components/BigNumber/BigNumber.tsx
```diff
@@ -40,10 +38,25 @@ export function BigNumber({
...
+	if (method === 'last') {
+		bigNumberValue = data && data.length > 0 ? data[data.length - 1][dataKey] : undefined;
+	} else if (method === 'sum' || method === 'avg') {
+		bigNumberValue = data && data.length > 0 ? data.reduce((sum, cur) => {
+			return sum + cur[dataKey];
+		}, 0) : undefined;
+		if (method === 'avg') {
+			bigNumberValue = bigNumberValue ? bigNumberValue / data.length : bigNumberValue;
+		}
+	}
```
gideontonkinson, line 59: I would probably check for data and data.length before in maybe a higher if statement. I like the way you reused sum to calc avg. though

src/components/BigNumber/BigNumber.tsx
```diff
@@ -95,42 +108,30 @@ export function BigNumber({
...
+				<View gridArea="icon">{iconElements}</View>
+				<View gridArea="data" UNSAFE_className="BigNumber-data">
+					{formattedValue}
+				</View>
+				<View gridArea="label" UNSAFE_className="BigNumber-label">
+					{label}
+				</View>
+			</Grid>
+		</div>
+	);
```
gideontonkinson, line 136: Great simplification here

src/stories/components/BigNumber/BigNumber.story.tsx
```diff
@@ -38,18 +38,71 @@ const BigNumberStory: StoryFn<typeof BigNumber> = (args): ReactElement => {
...
+		width: 600,
+		height: 600,
+	});
+	return (
+		<Chart {...chartProps}>
+			<BigNumber {...args}/>
+		</Chart>
+	);
+};
+
```
gideontonkinson, line 99: Nice additional coverage here

src/stories/components/BigNumber/BigNumber.story.tsx
gideontonkinson: Yeah we can remove both of these stories, my bad though as I had commented them out.

src/stories/components/BigNumber/BigNumber.test.tsx
gideontonkinson: Retweet

gideontonkinson (changes requested):
Overall great work, just a few changes here that I think will improve the code, the main one being the default value on method.

src/stories/components/BigNumber/BigNumber.story.tsx
Judd2000: OK - in that case, since data can't ever be null as it is  a required property of Chart - do you all think we need it?

jacob-hart:
We have decent test coverage so I'd like to get this merged in.  I can work on code smells and getting up to 100% coverage

PR #126, opened 2024-02-13
Add BigNumber component

gideontonkinson:
## Description

Add BigNumber component to RSC components
BigNumber is a child of Chart
BigNumber can accept a single line as a child
BigNumber displays the sum, avg or last value of the data set (determined by user-defined props) of the chart

## Motivation and Context

BigNumber allows for a user to display a number and a label along with a sparkline to help emphasize a specific point of data

## How Has This Been Tested?

There is 100% test coverage on the BigNumber code and tests of all potential layouts and number formats

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [X ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


jacob-hart:
I've been working on the docs for this component and think this code needs prop defaults on `numberType` and `method` (possibly more props too).  That way we can remove some unnecessary coalescing code

src/stories/components/BigNumber/BigNumber.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+		test('Vertical orientation sparkline', async () => {
+			render(<SparklineVertical {...SparklineVertical.args} />);
+			const value = await screen.findByText('20');
+			expect(value).toBeInTheDocument();
+
+			const label = await screen.findByText('Visitors');
+			expect(label).toBeInTheDocument();
+
+			const chart = await findChart();
+			expect(chart).toBeInTheDocument();
```
c-lamoureux, line 159: We should add an assertion for the line as well so we can be sure that it's being included in the chart.

src/stories/components/BigNumber/BigNumber.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+		test('Horizontal orientation sparkline', async () => {
+			render(<SparklineHorizontal {...SparklineHorizontal.args} />);
+			const value = await screen.findByText('20');
+			expect(value).toBeInTheDocument();
+
+			const label = await screen.findByText('Visitors');
+			expect(label).toBeInTheDocument();
+
+			const chart = await findChart();
+			expect(chart).toBeInTheDocument();
```
c-lamoureux, line 170: Same as above. We should check for the line.

src/utils/utils.ts
```diff
@@ -57,6 +58,21 @@ export const sanitizeRscChartChildren = (children: Children<RscElement> | undefi
...
 
+export const sanitizeBigNumberChildren = (children: Children<BigNumberChildElement> | undefined): BigNumberChildElement[] => {
+	const sanitizedChildren = toArray(children)
+		.flat()
+		.filter((child): child is BigNumberChildElement => isBigNumberChildElement(child));
+	return sanitizedChildren.filter(c => c.type == Line)
+};
+
+export const chartContainsBigNumber = (children: Children<ChartChildElement> | undefined): ChartChildElement[] => {
+	console.warn('Chart contains big number out of these children', children);
```
c-lamoureux, line 69: Let's drop this warning since it's expected behavior.

src/utils/utils.ts
```diff
@@ -57,6 +58,21 @@ export const sanitizeRscChartChildren = (children: Children<RscElement> | undefi
...
+	const sanitizedChildren = sanitizeRscChartChildren(children);
+	console.warn('After children have been sanitized', sanitizedChildren);
```
c-lamoureux, line 71: Same as above.

src/utils/utils.ts
```diff
@@ -85,8 +101,15 @@ export const sanitizeTrendlineChildren = (
 };
 
 const isChartChildElement = (child: ChildElement<ChartChildElement> | undefined): child is ChartChildElement => {
-	return isRscComponent(child);
+	const res = isRscComponent(child);
+	console.warn('Is the child an rsc component?', res);
+	return res;
```
c-lamoureux, line 106: It looks like this might have been for debugging. Lets revert it back.

src/utils/utils.ts
```diff
@@ -85,8 +101,15 @@ export const sanitizeTrendlineChildren = (
...
 };
+
+const isBigNumberChildElement = (child: ChildElement<BigNumberChildElement> | undefined): child is BigNumberChildElement => {
```
c-lamoureux, line 109: We can replace the usage of this function with `isChartChildElement` since their implementation is the same.

src/Chart.tsx
```diff
@@ -104,6 +110,49 @@ export const Chart = forwardRef<ChartHandle, ChartProps>(
 			);
 		}
 
+		if (hasBigNumberChildren && toArray(props.children).length != bigNumberChildren.length) {
+			throw new Error(
+				'If passing BigNumber to Chart only the BigNumber will be displayed. All other elements will be ignored'
```
c-lamoureux, line 115: Ideally we could throw an error during development if the children being passed are incompatible. That's most likely pretty complicated to put together though. Even though it's best practice to avoid silent failures, I think it would be preferable over having the API allow the users to pass in multiple children only to throw an error. 

Let's make this a console.warn instead.

c-lamoureux (changes requested):

src/components/BigNumber/BigNumber.css
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2023 Adobe. All rights reserved.
```
marshallpete, line 2: All these copyrights should say 2024 since this code is being added in 2024

src/components/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,247 @@
+/* eslint-disable react-hooks/rules-of-hooks */
+
+/*
+ * Copyright 2023 Adobe. All rights reserved.
```
marshallpete, line 4: 2024

src/components/BigNumber/index.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright 2023 Adobe. All rights reserved.
```
marshallpete, line 2: 2024

src/stories/components/BigNumber/BigNumber.story.tsx
```diff
@@ -0,0 +1,223 @@
+/* eslint-disable react-hooks/rules-of-hooks */
+
+/*
+ * Copyright 2023 Adobe. All rights reserved.
```
marshallpete, line 4: 2024

src/components/BigNumber/bigNumberFormatUtils.ts
```diff
@@ -0,0 +1,25 @@
+import { BigNumberProps } from 'types';
```
marshallpete, line 1: This needs the copyright

run `yarn copyrights` from the root folder and the script will add it to all files that need it.

src/components/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+	const { areas, columns, iconAlign, labelAlign, iconJustify, iconSize, displayCombo } = getGridProperties(
+		icon,
+		orientation,
+		height,
+		chartWidth,
+		lineProps
+	);
+	return (
+		<div tabIndex={0} className="big-number-container">
```
c-lamoureux, line 90: Let's just drop the tabIndex for now. We don't have accessibility support yet, so we can come back to this when we do a larger push for it in the future.

src/components/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,273 @@
...
+	orientation: Orientation,
+	chartWidth: number,
+	hasIcon: boolean,
+	hasLines: boolean,
+	height?: number
+): 'XXS' | 'XS' | 'S' | 'M' | 'L' | 'XL' | 'XXL' {
+	if (hasIcon) {
+		if (hasLines) {
+			if (orientation == 'vertical') {
+				return determineIconSize(height ? height / 3 : chartWidth / 2);
```
c-lamoureux, line 233: pulling this ternary up into a variable would reduce the Sonar Cloud cognitive complexity.
```suggestion
		if (hasLines) {
		        const availableWidth = height ? height / 3 : chartWidth / 2
			if (orientation == 'vertical') {
				return determineIconSize(availableWidth);
```

src/components/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,273 @@
...
+			}
+			if (orientation == 'horizontal') {
+				return determineIconSize(chartWidth / 6);
+			}
+		}
+
+		if (orientation == 'vertical') {
+			return determineIconSize(height ? height / 1.75 : chartWidth);
```
c-lamoureux, line 241: We could pull this one up as well for consistency but I don't think it will help with the complexity score.

src/components/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,273 @@
...
+}
+
+function getIconSizeByOrientation(
+	orientation: Orientation,
+	chartWidth: number,
+	hasIcon: boolean,
+	hasLines: boolean,
+	height?: number
+): 'XXS' | 'XS' | 'S' | 'M' | 'L' | 'XL' | 'XXL' {
+	if (hasIcon) {
```
c-lamoureux, line 230: We could make this the default check instead to reduce nesting:
```suggestion
	if (!hasIcon) return 'L';
	
	if (hasLines) { ...
```

src/components/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,273 @@
...
+				};
+			}
+		} else {
+			if (orientation === 'vertical') {
+				return { areas: ['sparkline', 'data', 'label'], columns: ['1fr'], labelAlign };
+			}
+			if (orientation === 'horizontal') {
+				return { areas: ['sparkline data', 'sparkline label'], columns: ['1fr 1fr'], labelAlign };
+			}
+		}
```
c-lamoureux, line 196: I would just extract this into a separate function since it's adding a lot to the complexity of the function.

src/components/BigNumber/BigNumber.tsx
```diff
@@ -0,0 +1,273 @@
...
+			iconJustify = 'end';
+			return {
+				areas: ['icon data', 'icon label'],
+				columns: ['1fr', '2fr'],
+				iconSize,
+				iconJustify,
+				labelAlign,
+			};
+		}
+	}
```
c-lamoureux, line 218: You could extract this one as well for consistency.

jacob-hart:
@c-lamoureux I've simplified the number formatting in `bigNumberFormatUtils`.  It does less by default now, which makes it more flexible for various locales and number formats.  I also added tests specifically for that file to make sure it handles the unhappy paths appropriately

src/components/BigNumber/BigNumber.tsx
```diff
@@ -49,209 +49,189 @@ const BigNumber: FC<BigNumberProps> = ({
...
-		if (height && height < chartWidth / (1.75 * aspectRatio)) {
-			cHeight = height;
-			cWidth = height * aspectRatio;
-		} else {
-			cWidth = chartWidth / 1.75;
-			cHeight = cWidth / aspectRatio;
-		}
-		padding = 10;
-		generalJustify = 'start';
+	let lineProps;
```
c-lamoureux, line 57: We want to avoid `any` types as much as possible. We should be able to type this:
```suggestion
	let lineProps: LineProps | undefined;
```

src/components/BigNumber/BigNumber.tsx
```diff
@@ -49,209 +49,189 @@ const BigNumber: FC<BigNumberProps> = ({
...
+	if (lineElements.length > 0) {
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
c-lamoureux, line 59: It looks like these are all being used. Is this lint disable necessary?

src/components/BigNumber/BigNumber.tsx
```diff
@@ -49,209 +49,189 @@ const BigNumber: FC<BigNumberProps> = ({
...
+	iconSize: BigNumberIconSize;
+	labelSize;
+	valueSize;
+	cHeight;
+	cWidth;
+	padding;
+	textAlign;
+	direction;
+	iconDirection;
+	pointSize;
```
c-lamoureux, line 146: Can we add types to these?

src/components/BigNumber/BigNumber.tsx
```diff
@@ -49,209 +49,189 @@ const BigNumber: FC<BigNumberProps> = ({
...
-			}
-		} else {
-			if (orientation === 'vertical') {
-				return { areas: ['sparkline', 'data', 'label'], columns: ['1fr'], labelAlign };
-			}
-			if (orientation === 'horizontal') {
-				return { areas: ['sparkline data', 'sparkline label'], columns: ['1fr 1fr'], labelAlign };
-			}
-		}
+	const iconSize = getDynamicIconSize(orientation, chartWidth, lineProps, height);
```
c-lamoureux, line 150: Great job breaking this out into multiple functions to reduce complexity!

PR #141, opened 2024-02-20
chore: add constants for scale names

marshallpete:
## Description

Just adding constants for the scale names to help avoid typos and make it easy to change in the future.

### DIFF
--- a/src/specBuilder/area/areaUtils.test.ts
+++ b/src/specBuilder/area/areaUtils.test.ts
@@ -9,7 +9,13 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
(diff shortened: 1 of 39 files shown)


src/specBuilder/area/areaUtils.test.ts
```diff
@@ -18,7 +24,7 @@ describe('getAreaMark', () => {
 		expect(
 			getAreaMark({
 				name: 'area0',
-				color: 'color',
```
c-lamoureux, line 21: There are a few places where we swapped 'color' with DEFAULT_COLOR. default color is 'series'. Just want to be sure that's intentional since this is intended as a refactor specifically.

c-lamoureux (dismissed):

PR #153, opened 2024-02-23
Nathaniel judd/byu team/animations

Judd2000:


src/VegaChart.tsx
```diff
@@ -69,9 +71,23 @@ export const VegaChart: FC<VegaChartProps> = ({
 		return { [TABLE]: clonedData };
 	}, [data]);
 
-	const previousChartData = usePreviousChartData({ [PREVIOUS_TABLE]: chartData.table });
+	const previousChartData = useMemo(() => {
+		const clonedData = JSON.parse(JSON.stringify(previousData));
```
jacob-hart, line 75: Could we use a true deep copy here?  We can import `structuredClone` from elsewhere in the project

src/VegaChart.tsx
```diff
@@ -119,9 +136,10 @@ export const VegaChart: FC<VegaChartProps> = ({
 		};
 	}, [
 		chartData.table,
-		previousChartData,
+		// previousChartData,
```
jacob-hart, line 139: We should be able to get the linter happy on this

src/specBuilder/specUtils.ts
```diff
@@ -296,4 +296,21 @@ export const usePreviousChartData = <T>(data: T) => {
...
+};
+
+export const getAnimationMarks = (dimension: string, metric: string, data?: ChartData[], previousData?: ChartData[]) => {
+	let markUpdate = {
+		scale: "yLinear",
+		signal: `datum.${metric} * timerValue`
+	};
+	if (data && previousData) {
+		const isCongruent = data.every((d) => previousData.some((pd) => d[dimension] === pd[dimension]));
+		if (isCongruent) {
```
jacob-hart, line 308: It would be good to put a comment describing what happens in the false condition.  I assume we ignore animations in that case?

src/specBuilder/specUtils.ts
```diff
@@ -296,4 +296,21 @@ export const usePreviousChartData = <T>(data: T) => {
...
\ No newline at end of file
+};
+
+export const getAnimationMarks = (dimension: string, metric: string, data?: ChartData[], previousData?: ChartData[]) => {
+	let markUpdate = {
+		scale: "yLinear",
+		signal: `datum.${metric} * timerValue`
+	};
+	if (data && previousData) {
+		const isCongruent = data.every((d) => previousData.some((pd) => d[dimension] === pd[dimension]));
```
jacob-hart, line 307: `isCongruent` is a nice short name, but I didn't immediately understand it.  Could we try `hasSameDimensions`?  Then it's more clear what we are branching on on the next line

src/stories/components/Line/Line.test.tsx
```diff
@@ -379,4 +380,20 @@ describe('Line', () => {
 			expect(point.getAttribute('stroke-width')).toEqual('2');
 		});
 	});
+
+	describe('Line animation tests', () => {
+		test('Line animates between congruent datasets', () => {
```
jacob-hart, line 385: Is this test running async or blocking?  We can probably use async to free up the test runner

jacob-hart (changes requested):
I'd like to get the linter happy before merging in.  You can do `yarn lint` locally to see what the complaints are

PR #168, opened 2024-03-05
Trendline display on hover refactor

marshallpete:
Switch from using `${name}_hoveredSeries` to using `highlightedSeries` everywhere.


tanugarg04 (approved):
looks great

PR #184, opened 2024-03-13
Animations disable animations for the dodged trellis bar

Judd2000:
Fixing more tests for animation implementation


jacob-hart (approved):

jacob-hart:
Correct me if this is wrong, but I think generally we should disable animations in the test code and not the stories.  We want the stories to reflect real use cases, which would almost always have the animations

PR #188, opened 2024-03-14
Adam/byu team/animations

agardn87:
## Description
This change adds opacity animations to line, bar, area, trendline, and legend. 

If a line, bar, or area chart has a tooltip, trendline, or popover, and animations are enabled, then upon hover, non selected marks are faded to transparency. Upon un-hover, non selected marks brighten back to opaque. If a legend is present, and all of the previous conditions are met, then upon hover the non-hovered legend symbols will fade to transparency, and upon un-hover the symbols will brighten back to opaque.
## Motivation and Context
Opacity animations allows for a cleaner user experience by slowing down the opacity changes upon hovering a chart. 

## How Has This Been Tested?
Tests have not been written yet. 

## Types of changes

## Checklist:


src/specBuilder/legend/legendSpecBuilder.test.ts
```diff
@@ -355,4 +377,43 @@ describe('addSignals()', () => {
 			)
 		).toBeUndefined();
 	});
+
+	test('should update colorAnimationDirection signal if animations are true', () => {
+		defaultSpec.signals = new Array<Signal>;
+		defaultSpec.signals.push(...defaultAnimationsSignals);
+		expect(addLegend(defaultSpec, { ...defaultLegendProps, animations: true }).signals)
+			.toStrictEqual([
+				{
```
jacob-hart, line 386: We could make this test more readable by extracting the expected object into a constant in the top of the file

src/specBuilder/legend/legendUtils.ts
```diff
@@ -222,6 +231,31 @@ const getHighlightOpacityEncoding = (
 		defaultOpacity,
 	];
 };
+//TODO: add comments/tests/etc
+const getHighlightOpAnimationValue = (opacityEncoding: BaseValueRef<number>,): { signal: string } => {
```
jacob-hart, line 235: Using `Op` in the name seems like a code smell that this function isn't doing enough on top of `getHighlightOpacityAnimationValue`.  We could use an import alias to avoid the name collision (OK), or (probably better) collapse the `if ('signal' in opacityEncoding)` logic into `getHighlightOpacityAnimationValue`

src/specBuilder/scale/scaleSpecBuilder.ts
```diff
@@ -200,3 +200,36 @@ const isScaleMultiFieldsRef = (
 ): domain is ScaleMultiFieldsRef => {
 	return Boolean(domain && !Array.isArray(domain) && 'data' in domain && 'fields' in domain);
 };
+
+export const getRSCAnimationScales = (scales: Scale[]) => {
```
jacob-hart, line 204: I think `getRSCAnimationScales` can mislead a reader, instead can we try `addRSCAnimationScales`

src/specBuilder/signal/signalSpecBuilder.ts
```diff
@@ -55,26 +55,29 @@ export const getSeriesHoveredSignal = (name: string, nestedDatum?: boolean, even
...
 	const hoveredSeries = `domain("${name}Entries")[datum.index]`;
 	const update = includeHiddenSeries
 		? `indexof(hiddenSeries, ${hoveredSeries}) === -1 ? ${hoveredSeries} : ""`
 		: hoveredSeries;
 	return {
-		name: keys ? `${name}_highlight` : 'highlightedSeries',
+		name: keys ? `${name}_highlight` : isPrev == true ? 'highlightedSeries_prev': `highlightedSeries`,
 		value: null,
-		on: [
+		on: !isPrev ? [
```
jacob-hart, line 75: It took me a bit to read this conditional.  There's no harm in building it in a `const` above

src/specBuilder/signal/signalSpecBuilder.ts
```diff
@@ -94,3 +97,95 @@ export const getLegendLabelsSeriesSignal = (value: unknown = null): Signal => {
 export const getGenericSignal = (name: string, value: unknown = null): Signal => {
 	return { name, value };
 };
+
+//TODO: Add documentation
+export const getAnimationSignals = (name: string): Signal[] => {
```
jacob-hart, line 102: good way to group these cleanly

src/specBuilder/signal/signalSpecBuilder.ts
```diff
@@ -94,3 +97,95 @@ export const getLegendLabelsSeriesSignal = (value: unknown = null): Signal => {
...
+		on: [{
+			events: 'timer{16}',
+			update: 'scale(\'rscAnimationCurve\', scale(\'rscAnimationCurveInverse\', rscAnimation) + 0.03333333333333334)'
+		}]
+	};
+};
+
+//TODO: add documentation
+const getRSCColorAnimationDirection = (name: string): Signal => {
+	if (name === 'line0') {
```
jacob-hart, line 133: Using `name` as a variable here can be confusing, as it visually conflicts with the object key called `name`

src/specBuilder/signal/signalSpecBuilder.ts
```diff
@@ -94,3 +97,95 @@ export const getLegendLabelsSeriesSignal = (value: unknown = null): Signal => {
 export const getGenericSignal = (name: string, value: unknown = null): Signal => {
 	return { name, value };
 };
+
+//TODO: Add documentation
+export const getAnimationSignals = (name: string): Signal[] => {
+	return [
+		getRSCAnimation(),
```
jacob-hart, line 104: We should use `Rsc` in all these function names.  Using an acronym is an edge case for CamelCase and the format can go either way.  But, because the project already uses `RscChart` future developers get a lot of value from us keeping consistent to `Rsc`

src/stories/components/Bar/StackedBar.story.tsx
```diff
@@ -35,7 +35,7 @@ const colors: SpectrumColor[] = [
 const BarStory: StoryFn<typeof Bar> = (args): ReactElement => {
 	const chartProps = useChartProps({ data: barSeriesData, colors, width: 800, height: 600 });
 	return (
-		<Chart {...chartProps}>
+		<Chart {...chartProps} debug>
```
jacob-hart, line 38: This is probably fine for now, but we need to make sure debug doesn't slip into the main branch.  One tip is to manually `git add` files that you see `git status`.  That way you only ever add files with feature changes, not debug logs/flags

jacob-hart (changes requested):
I left some minor readability/naming comments.  Before merging, we should make sure these commits line up on top of the existing `animations` branch.  That might mean copying these changes over to new commits

PR #204, opened 2024-03-26
Animations fix popover bug and add animations off capabilities to some stories where it is necessary

Judd2000:


jacob-hart:
It looks like most of the changes besides `RscChart.tsx` are just concurrent changes from main branch.  So I'll just focus on reviewing `RscChart.tsx`

src/RscChart.tsx
```diff
@@ -104,14 +106,24 @@ export const RscChart = forwardRef<ChartHandle, RscChartProps>(
...
 
-		// THE MAGIC, builds our spec
+		const [rerender, setRerender] = useState<boolean>(true);
+
+		const invertRerender = useRef(() => setRerender(!rerender));
+
+		useEffect(() => {
+			doAnimateFromZero.current = true;
+			// ensure a rerender so the current chart animates.
+			invertRerender.current();
```
jacob-hart, line 116: .This is pretty funky!  I wouldn't expect that setting state in a ref would be different from just setting state directly in the useEffect.  But if those code does the job, it might be OK

src/RscChart.tsx
```diff
@@ -163,7 +178,9 @@ export const RscChart = forwardRef<ChartHandle, RscChartProps>(
 			padding
 		);
 
-		const tooltipConfig: TooltipOptions = { theme: colorScheme };
```
jacob-hart, line 166: Could we have a comment here for why the `useMemo`?  Correct me if I'm wrong, but I think colorScheme is just a prop we pass down, so we shouldn't need to cache it here

src/RscChart.tsx
```diff
@@ -95,6 +95,8 @@ export const RscChart = forwardRef<ChartHandle, RscChartProps>(
 		forwardedRef
 	) => {
 		// uuid is used to make a unique id so there aren't duplicate ids if there is more than one Chart component in the document
+		const doAnimateFromZero = useRef(true);
```
jacob-hart, line 98: Correct me if this is wrong, but I think this value should be state instead?  The boolean value probably affects the render result so it can be counterintuitive to keep it as a ref.

You probably already know this, but React processes state changes in a batch.  I don't know how it interops with refs though

jacob-hart (commented):
I'd like to see a bit more comments explaining the hooks.  It is pretty complex and mingles refs and state, so we want to make sure future readers can understand the interactions.  Also it would be nice to get the linter happy

src/RscChart.tsx
Judd2000: You were right - I think the ref was an artifact from a previous fix I was trying! It works (and looks) much better as a state variable, too.

PR #207, opened 2024-03-27
Njudd/byu team/animations - fixes to animation function when popover pressed, fixing trellis bar and other tests broken by animations

Judd2000:
Some of @gideontonkinson's changes are here to use with testing, focus on the changes made to RscChart.tsx, typings with animateFromZero added, and existing tests.


src/RscChart.tsx
```diff
@@ -104,14 +107,20 @@ export const RscChart = forwardRef<ChartHandle, RscChartProps>(
 
 		const sanitizedChildren = sanitizeRscChartChildren(props.children);
 
-		// THE MAGIC, builds our spec
+		// when data changes, make sure that we are animating from zero (especially in the case where a popover was just
+		// opened and closed)
+		useEffect(() => {
+			setAnimateFromZero(true);
+		}, [data]);
```
jacob-hart, line 114: Correct me if this is wrong, but I thought we were only wanting to animate from zero if data was too dissimilar to the previous data.  Is this accounted for later on in the code?

src/RscChart.tsx
```diff
@@ -137,6 +146,8 @@ export const RscChart = forwardRef<ChartHandle, RscChartProps>(
 			// if the popover is closed, reset the selected data
 			if (!popoverIsOpen) {
```
jacob-hart, line 147: nit: this condition might be more intuitive if we swapped the statements.  Right now I read it like "if the popover isn't open (i.e. it closed)" ,  rather than my preferred reading of "if the popover just opened"

jacob-hart (commented):
I left comments with some minor nits, but overall if we can get it to build and pass tests we can merge it in

src/RscChart.tsx
Judd2000: Correct, it is!

PR #213, opened 2024-04-01
Animations made area and line charts have tooltips and highlight mark…

Judd2000:


src/specBuilder/area/areaUtils.ts
```diff
@@ -76,20 +76,21 @@ export const getAreaMark = (
 	interactive: getInteractive(children),
 	encode: {
 		enter: {
-			...((animations === false || !animateFromZero) && {
+			...((!animations || !animateFromZero) && {
```
jacob-hart, line 79: Treating `animations` as a pure boolean makes it more readable here, good job

src/specBuilder/marks/markUtils.ts
```diff
@@ -77,10 +77,25 @@ export function getInteractive(children: MarkChildElement[]): boolean {
 /**
  * If a tooltip or popover exists on the mark, then set tooltip to true.
  */
-export function getTooltip(children: MarkChildElement[], name: string, nestedDatum?: boolean): SignalRef | undefined {
+export function getTooltip(
```
jacob-hart, line 80: This seems like a funky interface for getTooltip, but I see how it makes sense to toggle the booleans by name.  I like how python does this using `kwargs` but I don't think js has a comparable feature, besides this approach

src/specBuilder/signal/signalSpecBuilder.ts
```diff
@@ -79,9 +79,29 @@ export const getGenericSignal = (name: string, value: unknown = null): Signal =>
...
+		signals: Signal[],
+		markName: string,
+		datumOrder?: number,
+		animations?: boolean,
+		animateFromZero?: boolean,
+		needsDisable?: boolean
+	}
+) => {
 	const highlightedItemSignal = signals.find((signal) => signal.name === HIGHLIGHTED_ITEM);
+	const highlightedUpdate = `${new Array(datumOrder).fill('datum.').join('')}${MARK_ID}`;
```
jacob-hart, line 104: The `new Array` constructor is pretty uncommon but I see how this is copy-pasted from below.  In this case, it looks like a good choice with the `fill-join` chain

src/stories/components/ChartTooltip/ChartTooltip.story.tsx
```diff
@@ -73,7 +73,7 @@ const lineData = [
 const LineTooltipStory: StoryFn<typeof ChartTooltip> = (args): ReactElement => {
 	const chartProps = useChartProps({ data: lineData, width: 600 });
 	return (
-		<Chart {...chartProps}>
+		<Chart {...chartProps} animations={false}>
```
jacob-hart, line 76: nit: let's be careful we don't let these slip into the storybook unless necessary

src/stories/components/Line/Line.story.tsx
```diff
@@ -141,7 +141,7 @@ const HistoricalCompareStory: StoryFn<typeof Line> = (args): ReactElement => {
 const LineWithVisiblePointsStory: StoryFn<typeof Line> = (args): ReactElement => {
 	const chartProps = useChartProps({ ...defaultChartProps, data: workspaceTrendsDataWithVisiblePoints });
 	return (
-		<Chart {...chartProps}>
+		<Chart {...chartProps} animations={false}>
```
jacob-hart, line 144: nit: let's be careful we don't let these slip into the storybook unless necessary

jacob-hart (approved):

PR #215, opened 2024-04-02
Ability to disable tooltip for chart data based on key

olsonjb:
## Description

- `excludeDataKey` prop added to `ChartTooltip`
-  Story demonstrating `excludeDataKey` use
-  Disable interaction, voronoi calculation, and tooltip display for data points where `excludeDataKey` is truthy

## Related Issue

[AN-344693](https://jira.corp.adobe.com/browse/AN-344693) (Internal)

## Types of changes

## Checklist:

### DIFF
--- a/src/components/ChartTooltip/ChartTooltip.tsx
+++ b/src/components/ChartTooltip/ChartTooltip.tsx
@@ -14,7 +14,7 @@ import { FC } from 'react';
 import { ChartTooltipProps } from '../../types';
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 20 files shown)


c-lamoureux:
The implementation for this looks good. I think `excludeDataKey` should probably be an array rather than a single key. It seems likely that it would be common for users to want to exclude multiple keys. 

I also noticed that there is some odd behavior with hovering when we exclude a data key. We should see if we can find a way to make it so the other line doesn't trigger a hover when there are excluded keys. 


https://github.com/adobe/react-spectrum-charts/assets/29240999/c45379ae-ae9b-4b0b-896c-b7f6ca06ca84

c-lamoureux (approved):

PR #244, opened 2024-04-15
feat: adds support for percentage based height

carrickjason:
## Description

- Support percentage based heights
- Support min/max heights
- Percentage based heights will be updated on container resize

## Related Issue

Closes #216 

## Motivation and Context

Allows Charts to fill all or a percentage of the available vertical space

## How Has This Been Tested?

- Tested with new story in storybook
- linked package to my team's implementation and verified it works with quarry/dashboard widgets

## Screenshots (if appropriate):

https://github.com/adobe/react-spectrum-charts/assets/12468946/d65d38c5-7644-421e-a3e0-c4a9f650ba94

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #248, opened 2024-04-16
feat: adds support for percentage based height

evargast:
See [PR-244](https://github.com/adobe/react-spectrum-charts/pull/244) for details

### DIFF
--- a/src/Chart.tsx
+++ b/src/Chart.tsx
@@ -121,49 +121,45 @@ export const Chart = forwardRef<ChartHandle, ChartProps>(
 				UNSAFE_style={{ backgroundColor: 'transparent' }}
 				height="100%"
 			>
-				<div
-					ref={containerRef}
-					id={chartId.current}
-					data-testid={dataTestId}
-					className="rsc-container"
-					style={{ backgroundColor: getColorValue(backgroundColor, colorScheme), height: '100%' }}
-				>
-					{showPlaceholderContent ? (
-						<PlaceholderContent
-							loading={loading}
-							data={data}
-							height={chartHeight}
-							emptyStateText={emptyStateText}
-						/>
-					) : (
-						<RscChart
-							chartView={chartView}
-							chartId={chartId}
-							data={data}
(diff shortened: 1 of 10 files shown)


src/Chart.tsx
```diff
@@ -74,14 +77,21 @@ export const Chart = forwardRef<ChartHandle, ChartProps>(
...
 		const chartView = useRef<View>();
 		const [containerWidth, setContainerWidth] = useState<number>(0);
+		const [containerHeight, setContainerHeight] = useState<number>(0);
 
 		useChartImperativeHandle(forwardedRef, { chartView, title });
 
 		const containerRef = useResizeObserver<HTMLDivElement>((_target, entry) => {
-			if (typeof width === 'number') return;
-			setContainerWidth(entry.contentRect.width);
+			if (typeof width !== 'number') {
```
marshallpete, line 85:
```suggestion

			if (typeof width !== 'number') {
```

src/Chart.tsx
```diff
@@ -74,14 +77,22 @@ export const Chart = forwardRef<ChartHandle, ChartProps>(
...
 		const [containerWidth, setContainerWidth] = useState<number>(0);
+		const [containerHeight, setContainerHeight] = useState<number>(0);
 
 		useChartImperativeHandle(forwardedRef, { chartView, title });
 
 		const containerRef = useResizeObserver<HTMLDivElement>((_target, entry) => {
-			if (typeof width === 'number') return;
-			setContainerWidth(entry.contentRect.width);
+
+			if (typeof width !== 'number') {
```
marshallpete, line 86:
```suggestion
			if (typeof width !== 'number') {
```

src/Chart.tsx
marshallpete, line 85:
```suggestion

			if (typeof width !== 'number') {
```

src/Chart.tsx
marshallpete, line 86:
```suggestion
			if (typeof width !== 'number') {
```

PR #275, opened 2024-05-01
build: enhancements for parcel

marshallpete:
## Description

typeVersions are used to get the correct types using most bundlers. These don't seem to work with Parcel so we also include export subpaths with conditional exports which do work with parcel.

We need both.


package.json
```diff
@@ -6,8 +6,16 @@
...
+		".": {
+			"import": "./dist/index.js",
+			"types": "./dist/@types/index.d.ts",
+			"default": "./dist/index.js"
+		},
+		"./alpha": {
+			"import": "./dist/alpha.js",
+			"types": "./dist/@types/alpha/index.d.ts",
+			"default": "./dist/alpha.js"
+		}
```
marshallpete, line 18: Parcel needs these so that it can find the right type files.

tsconfig.json
```diff
@@ -2,14 +2,14 @@
...
+		"module": "node16",
 		"target": "ES2017",
 		"jsx": "react-jsx",
 		"lib": ["ESNext", "ES5", "ES6", "DOM", "DOM.Iterable"],
 		"sourceMap": true,
 		"strict": true,
 		"allowJs": false,
-		"moduleResolution": "node",
+		"moduleResolution": "node16",
 		"forceConsistentCasingInFileNames": true,
```
marshallpete, line 13: ts documentation highly recommends that libraries use node16 or nodenext. This does not mean that our library has to use v16 of node.

c-lamoureux (approved):

PR #276, opened 2024-05-01
Build updates

marshallpete:
## Description

Clean up imports since libraries shouldn't use named imports in files that get exported.


.eslintrc
```diff
@@ -9,6 +9,13 @@
...
 		"react/jsx-uses-vars": "error",
-		"react/jsx-uses-react": "error"
+		"react/jsx-uses-react": "error",
+		"no-restricted-imports": [
+			"error",
+			{
+				"name": "types",
+				"message": "Please use relative path import for types instead (ex. ../types)."
+			}
+		]
```
marshallpete, line 19: eslint rule to error is someone tries to import `from 'types'`

PR #292, opened 2024-05-08
chore: clean up the legend highlight logic

marshallpete:
A while ago we updated the highlight logic so that `opacity` is used to highlight and `stroke-opacity` and `fill-opacity` are used to set the static opacity values. This update didn't make it all the way down to legend highlight utils. This PR gets that up to date and a couple other house cleaning items.


tanugarg04 (approved):
LGTM

Issue #364, opened 2024-07-12
New Combo component for combo charts


marshallpete:
### Provide a general summary of the feature here

Combo charts are useful for displaying multiple types of data in the same visualization that you want to compare against each other.

Combo charts often have some unique edge cases like dual metric axes. They also need to share the same dimension axis. Combo charts also have unique highlight/hover behavior to consider.

A new component should be created that will wrap the combo chart types the user wants and handle these edge cases without any special configuration from the user.

To begin with, only a single bar and line combo will be supported.

### 🤔 Expected Behavior?

Dual axes should be possible
Should be possible to hover each bar and point on the line
Dimension axes should line up correctly across line and bar

## API
```
<Chart>
  <Axis position="right" name="people" />
  <Axis position="left" name="adoption" />
  <Axis position="bottom" />
  <Combo dimension="datetime">
    <Bar metric="people" metricAxis="people" />
    <Line metric="adoptionRate" metricAxis="adoption" />
    <ChartTooltip highlightBy="item"></ChartTooltip>
  </Combo>
</Chart>
```

Need to add `name` to the `Axis` component and be able to use that to correlate which mark uses which axis. `name` can default to the value of `position`.

Bar needs `metricAxis` added to it which will allow the bar metric to be connected to the correct axis. If undefined, the typical metric scale should be used.

New `Combo` component.
Should have a `dimension` prop which sets the dimension of the `Bar` and `Line`. If a dimension is set on either `Bar` or `Line`, this will override those values. Should default to something like `datetime`.

`ChartTooltip` needs a new `highlightBy` prop which will allow the user to configure the hover targets. Value of this prop can be `'item' | 'dimension' | 'series'`. We also want to be able to set some kind of group to highlight by but that can be handled in a separate ticket.

If `ChartTooltip` is nested under a `Combo` (child or grandchild) then for a line, this should use a hover area near the points on the line instead of `voronoi`. If `voronoi` is used then none of the bars are accessible.

### 🔦 Context

Combo charts are useful for displaying multiple types of data in the same visualization that you want to compare against each other.

### 🧢 Your Company/Team

Adobe

pratyushbanerjee:
Since we want the axes to behave differently for `Combo`, should the `Axis` components be children of `Combo`?

c-lamoureux:
[...]

PR #401, opened 2024-08-28
feat: Add item interaction mode for line

pratyushbanerjee:
## Description

Add item interaction mode for `Line` that triggers tooltip when hovering over a particular area near points on the line

## Related Issue

https://github.com/adobe/react-spectrum-charts/issues/364

## Motivation and Context

Voronoi method for tooltips on `Line` does not allow interaction with elements behind the line. `item` interaction mode allows for more precise interactions with the points on the `Line` and allows interaction with elements behind it.

## How Has This Been Tested?

Passed existing tests
Added tests to cover changes

## Types of changes

## Checklist:

### DIFF
--- a/src/specBuilder/chartTooltip/chartTooltipUtils.ts
+++ b/src/specBuilder/chartTooltip/chartTooltipUtils.ts
@@ -170,7 +170,7 @@ export const addTooltipSignals = (signals: Signal[], markProps: TooltipParentPro
 	}
 };
 
-function addMouseEvents(highlightedGroupSignal: Signal, markName: string, update: string) {
+const addMouseEvents = (highlightedGroupSignal: Signal, markName: string, update: string) => {
 	if (highlightedGroupSignal.on === undefined) {
 		highlightedGroupSignal.on = [];
 	}
@@ -183,7 +183,7 @@ function addMouseEvents(highlightedGroupSignal: Signal, markName: string, update
 			{ events: `@${markName}:mouseout`, update: 'null' },
 		]
 	);
-}
+};
 
 /**
  * adds the appropriate opacity rules to the beginning of the opacityRules array
--- a/src/specBuilder/chartTooltip/chartTooltipUtils.ts
+++ b/src/specBuilder/chartTooltip/chartTooltipUtils.ts
@@ -9,7 +9,6 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-
 import { ChartTooltip } from '@components/ChartTooltip';
 import {
 	DEFAULT_OPACITY_RULE,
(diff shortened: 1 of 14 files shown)


src/specBuilder/chartTooltip/chartTooltipUtils.ts
```diff
@@ -150,28 +151,40 @@ export const isHighlightedByDimension = (markProps: TooltipParentProps) => {
...
 	if (isHighlightedByGroup(markProps)) {
 		const highlightedGroupSignal = signals.find((signal) => signal.name === HIGHLIGHTED_GROUP) as Signal;
-		if (highlightedGroupSignal.on === undefined) {
-			highlightedGroupSignal.on = [];
-		}
 
 		let markName = markProps.name;
 		let update = `datum.${markName}_groupId`;
+
+		if ((markProps as LineSpecProps).interactionMode === INTERACTION_MODE.ITEM) {
```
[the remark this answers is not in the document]
marshallpete: It is better to check if `interactionMode` is on `markProps` than to cast `markProps` as `LineSpecProps`

src/specBuilder/line/lineMarkUtils.ts
```diff
@@ -152,3 +152,34 @@ const getHoverRule = (dimension: string, name: string, scaleType: ScaleType): Ru
...
+};
+
+const getItemHoverMarks = (dataSource: string, lineProps: LineMarkProps): Mark[] => {
+	const { children, dimension, metric, name, scaleType } = lineProps;
+
+	return [
+		// area around item that triggers hover
+		getItemHoverArea(children, dataSource, dimension, metric, name, scaleType),
+	];
+};
```
marshallpete, line 185: Love how you have broken these up.

marshallpete: jsdoc comments can be nice for these type of functions but aren't required.

src/specBuilder/chartTooltip/chartTooltipUtils.ts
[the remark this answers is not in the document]
pratyushbanerjee: Makes sense, more extensible

src/specBuilder/marks/markUtils.ts
```diff
@@ -338,6 +342,59 @@ export const getVoronoiPath = (children: MarkChildElement[], dataSource: string,
...
+					shape: { value: HOVER_SHAPE },
+					y: { scale: 'yLinear', field: metric },
+					fill: { value: 'transparent' },
+					stroke: { value: 'transparent' },
+					tooltip: getTooltip(children, name, false),
+					size: getHoverSizeSignal(size),
+				},
+				update: {
+					x: getXProductionRule(scaleType, dimension),
+				},
```
marshallpete, line 381:
```suggestion
				enter: {
					shape: { value: HOVER_SHAPE },
					y: { scale: 'yLinear', field: metric },
					fill: { value: 'transparent' },
					stroke: { value: 'transparent' },
					tooltip: getTooltip(children, name, false),
				},
				update: {
					x: getXProductionRule(scaleType, dimension),
					size: getHoverSizeSignal(size),
				},
```

marshallpete: Looks like size is width dependent. If something is width dependent, it needs to be in the `update`. The reason for this is width can change dynamically. For example let's say someone resizes the window. This may adjust the chart width if it is set as a percentage. If `size` is not in the `update`, it won't be recalculated when this happens.

src/specBuilder/line/lineMarkUtils.ts
marshallpete: Yes I agree. That's why I don't think you need to add it.

PR #515, opened 2025-02-15
Added support for horizontal alignment

rencherg:
Support for horizontal ordering of the bullet charts

## Description

Support for horizontal ordering of the bullet charts. A new direction prop was added so that the user can specify which ordering direction they want. Most changes happened in bulletSpecBuilder.ts. Unit tests were also updated.

## Related Issue

SCRUM-21
Use Case C

## Motivation and Context

This change is required to complete use case C of the Bullet Chart Use Cases

## How Has This Been Tested?

Unit tests were updated to support the new code

## Checklist:


src/specBuilder/bullet/bulletSpecBuilder.test.ts
```diff
@@ -10,11 +10,12 @@
  * governing permissions and limitations under the License.
  */
 
+// import { GroupMark } from 'vega';
```
marshallpete, line 13:
```suggestion
```

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -31,9 +31,11 @@ export const addBullet = (
         dimension,
         target,
         color = DEFAULT_COLOR,
+        direction,
```
marshallpete, line 34:
```suggestion
        direction = DEFAULT_BULLET_DIRECTION,
```

marshallpete: You will need to add this to the `constants.ts` file for this to work. This will default the value of direction to `vertical` if it isn't defined.

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -43,32 +45,41 @@ export const addBullet = (
         dimension: dimension ?? 'graphLabel',
         target: target ?? 'target',
         name: toCamelCase(name ?? `bullet${index}`),
+        //direction must be horizontal or vertical
+        direction: direction === 'vertical' || direction === 'horizontal' ? direction : 'vertical',
```
marshallpete, line 49:
```suggestion
        direction,
```

marshallpete: If you do the suggestion on line 34, you don't need to do this check.

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -43,32 +45,41 @@ export const addBullet = (
...
   
-  return [
+  const verticalMarks: Mark[] = [
     {
       "type": "rect",
-      "name": `${props.name}rect`,
+      "name": "bullet0rect",
       "from": {"data": "table"},
-      "description": `${props.name}`,
+      "description": "bullet0",
```
marshallpete, line 70: Any reason to go away from the dynamic name?

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -78,13 +89,17 @@ export function getBulletMarks(props: BulletSpecProps): Mark[] {
...
+      "name": "bullet0barlabel",
       "from": {"data": "table"},
       "description": "graphLabel",
       "encode": {
         "enter": {
           "x": {"value": 0},
-          "y": {"field": "index", "mult": 60, "offset": -60},
+          "y": {
+            "field": "index",
+            "mult": 60,
```
marshallpete, line 100: What does `mult` do? I'm not familiar.

marshallpete: Looks like it's a multiplier? Why do we need that here?

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -113,19 +132,129 @@ export function getBulletMarks(props: BulletSpecProps): Mark[] {
...
+      },
+      "scales": [
+        {
+          "name": "xscale",
+          "type": "linear",
+          "domain": [0, { "signal": "data('max_values')[0].maxOverall" }],
+          "range": [0, { "signal": "width" }],
+          "round": true,
+          "zero": true
+        }
```
marshallpete, line 187: Something like this is cleaner:
```        {
          "name": "xscale",
          "type": "linear",
          "domain": {
            "data": "table",
            "fields": ["xPaddingForTarget", "amount"]
          },
          "range": "width",
          "round": true,
          "zero": true
        }
```

Instead of passing in two values where you have to reference the data set and get the max overall, you can just tell it which fields to use and it will figure out the min and max for you.

This is what the original data looks like to get this to work:
```
 "data": [
    {
      "name": "table",
      "values": [
        { "category": "A", "amount": 10, "target": 40, "label": "Facebook" },
        { "category": "B", "amount": 65, "target": 40, "label": "Instagram" },
        { "category": "C", "amount": 33, "target": 40, "label": "Local" }
      ],
      "transform": [
        {
          "type": "formula",
          "expr": "round(datum.target * 1.05)",
          "as": "xPaddingForTarget"
        }
      ]
    }
  ],
```

marshallpete: If you do this, you don't need a `max_values` table at all. This can be a separate follow up PR. Just thought I'd share.

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -113,19 +132,129 @@ export function getBulletMarks(props: BulletSpecProps): Mark[] {
...
+              "x": { "scale": "xscale", "field": `${props.target}` },
+              "y": { "value": 12 }, 
+              "y2": { "value": 36 }, 
+              "stroke": { "value": `${solidColor}` },
+              "strokeWidth": { "value": 2 }
+            }
+          }
+        }
+      ]
+    }
```
marshallpete, line 254: This is a lot of lines of config for a single function. It would be nice to break this up into separate utility functions like `getLabelTextMark`, `getValueTextMark`, `getbulletTargetRule`, `getBulletMark`. These could then be called, eliminating 60+ lines of code from this one function. We prefer to keep all functions under 20 or so lines (not a hard and fast rule but a good guideline. It will make all of this a lot easier to read and understand what they each do.

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -113,19 +132,129 @@ export function getBulletMarks(props: BulletSpecProps): Mark[] {
...
+              "text": { "field": `${props.metric}` },
+              "align": { "value": "right" }, 
+              "baseline": { "value": "bottom" },
+              "fontSize": { "value": 11.5 },
+              "fill": { "value": `${solidColor}` }
+            }
+          }
+        },
+        {
+          "type": "rule",
```
marshallpete, line 240: Names like this should be dynamic and follow our mark naming convention. `bullet0_target` which is actually `` `${props.name}_target` ``

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -113,19 +132,129 @@ export function getBulletMarks(props: BulletSpecProps): Mark[] {
...
+              "text": {"field": `${props.dimension}`},
+              "align": {"value": "left"},
+              "baseline": {"value": "bottom"},
+              "fill": {"value": `${barLabelColor}`},
+              "fontSize": {"value": 11.5}
+            }
+          }
+        },
+        {
+          "type": "text",
```
marshallpete, line 224: `` `${props.name}_valueLabel` ``

marshallpete: Sorry this comment should be for line 226

src/specBuilder/bullet/bulletSpecBuilder.ts
marshallpete: Sorry this comment should be for line 242

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -113,19 +132,129 @@ export function getBulletMarks(props: BulletSpecProps): Mark[] {
...
+              "height": {"value": 6},
+              "fill": { "value": `${props.color}` },
+              "cornerRadiusTopRight": { "value": 2 },
+              "cornerRadiusBottomRight": { "value": 2 }
+            }
+          }
+        },
+        {
+          "type": "text",
+          "name": "bullet0barlabel",
```
marshallpete, line 208: `` `${props.name}_label` ``

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -113,19 +132,129 @@ export function getBulletMarks(props: BulletSpecProps): Mark[] {
...
+          "type": "linear",
+          "domain": [0, { "signal": "data('max_values')[0].maxOverall" }],
+          "range": [0, { "signal": "width" }],
+          "round": true,
+          "zero": true
+        }
+      ],
+      "marks": [
+        {
+          "name": "bullet0rect",
```
marshallpete, line 191: `` `${props.name}_bullet` ``

src/types/Chart.ts
```diff
@@ -237,6 +237,8 @@ export interface BulletProps extends MarkProps {
 	target?: string;
 	/** Data field that the metric is trended against (x-axis for horizontal orientation) */
 	dimension?: string;
+	/** Specifies the direction the bars should be ordered (vertical/horizontal) */
+	direction?: string;
```
marshallpete, line 241:
```suggestion
	direction?: 'row' | 'column';
```

marshallpete: This narrows that scope of allowable strings to only `row` or `column`. We want to use row/column, not vertical/horizontal.

marshallpete:
There are a lot of places that use `vertical` and `horizontal` but instead we would like to use `row` or `column`

src/specBuilder/bullet/bulletSpecBuilder.ts
rencherg: It multiplies the y position by the index, making it so every object in the data is placed lower than the one before it.

rencherg:
> There are a lot of places that use `vertical` and `horizontal` but instead we would like to use `row` or `column`

I should have changed them all

src/types/Chart.ts
rencherg: Fixed

src/specBuilder/bullet/bulletSpecBuilder.ts
rencherg: No, it should be fixed

src/specBuilder/bullet/bulletSpecBuilder.ts
rencherg: When row is specified we are doing the trellising to scale everything inside the alloted width but when column is specified we are using mult to give a set amount of pixels per bullet chart. I was hesitant to do the trellising for column mode because stretching/shrinking the bullet charts to fit would would not look great. Spacing out the bullets with mult works well but the only downside is that it gets cut off if the specified height is too small. We can change to trellising or different method if you prefer.

marshallpete: I see. Makes sense. We can discuss this on Monday. I have some ideas on how to simplify the row method so that it will make the chart grow/shrink but it doesn't need to be in the PR or anything.

rencherg: Sounds good

src/specBuilder/bullet/bulletSpecBuilder.ts
rencherg: All large functions have been broken up.

src/specBuilder/bullet/bulletMarkUtils.ts
```diff
@@ -0,0 +1,224 @@
...
+export function getBulletSignals(props: BulletSpecProps): Signal[] {
+    const bulletSignals: Signal[] = [
+        { "name": "gap", "value": 12 },
+        { "name": "bulletHeight", "value": 8 },
+        { "name": "bulletThresholdHeight", "update": "bulletHeight * 3" },
+        { "name": "targetHeight", "update": "bulletThresholdHeight + 6" },
+        { "name": "bulletGroupHeight", "update": "bulletThresholdHeight + 24" }
+    ]
+
+    if(props.direction === 'column'){
```
c-lamoureux, line 52: There are some style issues. We should make sure prettier is being applied to all modified files.

src/specBuilder/bullet/bulletMarkUtils.ts
```diff
@@ -0,0 +1,224 @@
...
+            "cornerRadiusBottomRight": [
+                { "test": `datum.${props.metric} > 0`, "value": 3 }
+            ],
+            "fill": [{ "value": `${props.color}` }]
+            },
+            "update": {
+                "x": { "scale": "xscale", "value": 0 },
+                "x2": { "scale": "xscale", "field": `${props.metric}` },
+                "height": { "signal": "bulletHeight" },
+                "y": { "signal": "bulletGroupHeight - 3 - 2 * bulletHeight" }
```
c-lamoureux, line 142: Can we add a comment explaining the context for these subtractions?

c-lamoureux, line 142: Can we add a comment explaining the context for these subtractions?

src/specBuilder/bullet/bulletSpecBuilder.test.ts
```diff
@@ -45,37 +59,127 @@ describe('addBullet', () => {
...
+//         expect(data[3].type).toBe('rule');
+//     });
+
+//     test('should return the correct row marks object', () => {
+//         const data = getBulletMarks(samplePropsRow) as GroupMark[];
+//         expect(data).toHaveLength(1);
+//         expect(data[0].type).toBe('group')
+//         expect(data[0].marks).toHaveLength(4);
+//     });
+// });
```
c-lamoureux, line 185: There's a lot of commented out code in this file. Was that intentional?

src/stories/components/Bullet/Bullet.story.tsx
```diff
@@ -24,18 +24,18 @@ export default {
...
+    height: 350,
 };
 
 // Basic Bullet chart story
 const BulletStory: StoryFn<BulletProps & { width?: number; height?: number }> = (args): ReactElement => {
     const { width, height, ...bulletProps } = args;
     const chartProps = useChartProps({ ...defaultChartProps, width: width ?? 350, height: height ?? 350 });
     return (
-        <Chart {...chartProps}>
+        <Chart {...chartProps} debug>
```
c-lamoureux, line 38:
```suggestion
        <Chart {...chartProps}>
```

src/stories/components/Bullet/Bullet.story.tsx
```diff
@@ -46,7 +46,8 @@ Basic.args = {
 	metric: 'currentAmount',
     dimension: 'graphLabel',
     target: 'target',
-    color: 'red-500'
+    color: 'red-500',
+    direction: 'column',
 };
 
 export { Basic };
```
c-lamoureux, line 53: We should add a story for the alignment now that it can be changed.

c-lamoureux:
It looks like the overall sizing of the bullet is a bit inconsistent between directions. I think that the width of the chart should stay the same when changing direction and the bullets should be sized accordingly within that width. Let me know if you have any questions about what I mean. Happy to review it further.

https://github.com/user-attachments/assets/3f297cec-949a-4877-a1ac-21e749dab7d5

src/specBuilder/bullet/bulletSpecBuilder.test.ts
rencherg: My bad, thanks for pointing it out

rencherg:
> It looks like the overall sizing of the bullet is a bit inconsistent between directions. I think that the width of the chart should stay the same when changing direction and the bullets should be sized accordingly within that width. Let me know if you have any questions about what I mean. Happy to review it further.
> 
>  Screen.Recording.2025-03-04.at.11.41.54.AM.mov

This should be fixed now

src/specBuilder/bullet/bulletMarkUtils.ts
rencherg: I tried to apply prettier and nothing was changed. Is there a specific Adobe style that I can import?

rencherg: Never mind, I got it to work for all new bullet files

src/stories/components/Bullet/Bullet.story.tsx
rencherg: Added

src/stories/components/Bullet/Bullet.story.tsx
```diff
@@ -35,7 +35,7 @@ const BulletStory: StoryFn<BulletProps & { width?: number; height?: number }> =
 	const { width, height, ...bulletProps } = args;
 	const chartProps = useChartProps({ ...defaultChartProps, width: width ?? 350, height: height ?? 350 });
 	return (
-		<Chart {...chartProps}>
+		<Chart {...chartProps} debug>
```
c-lamoureux: Extra debug here.

c-lamoureux (approved):

PR #520, opened 2025-02-25
Scrum 26/vertical group marks

rencherg:
Added better handling for multiple bullet charts in vertical orientation.

## Description

Changed the Marks field of the spec to include a MarkGroup, giving more control over how the scaling works.

## Related Issue

SCRUM-26

## Motivation and Context

The way the Marks field was set up was unstable and used a lesser common method of separating the bullet charts.

## How Has This Been Tested?

New tests were introduced to better test all elements of the getBulletMarks function and its helper functions

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+                    "as": "xPaddingForTarget"
+                }
+            ]
+        }
+    ]
+  
+    return bulletData;
+}
+
+export function getBulletMarks(props: BulletSpecProps): GroupMark[] {
```
spenjnb, line 76: This nice refactoring here. It is good to use the GroupMark to keep the code cleaner and scalable. It should make it easier to maintain and update.

src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+export function getBulletData(props: BulletSpecProps): Data[] {
+
+    //We are multiplying the target by 1.05 to make sure that the target line is never at the very end of the graph
+    const bulletData: Data[] = [
+        {
+            "name": "table",
+            "values": [],
+            "transform": [
```
spenjnb, line 63: Nice use of formula, window and maxValue this will be very helpful.

src/specBuilder/bullet/markUtils.test.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { getBulletScales, getBulletData, getBulletMarks, getBulletSignals, getBulletMarkRect, getBulletMarkLabel, getBulletMarkTarget, getBulletMarkValueLabel } from "./markUtils";
+import { sampleProps } from "./bulletSpecBuilder.test";
+
+describe('getBulletMarks', () => {
```
spenjnb, line 16: This is a good and detailed test case for mark types!

src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+            "name": "yscale",
+            "type": "band",
+            "domain": { "data": "table", "field": `${props.dimension}` },
+            "range": [0, { "signal": "height" }],
+            "paddingInner": { "signal": "paddingRatio" }
+        },
+        {
+            "name": "xscale",
+            "type": "linear",
+            "domain": { "data": "table", "fields": ["xPaddingForTarget", `${props.metric}`] },
```
spenjnb, line 29: Here I am thinking if there could be any case where xPaddingForTarget is greater that the metric (props.metric) what would happen to the scale this might be overreaching but only in a unexpected scenario where target is way larger than the actual metric.

src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+                    "y": { "scale": "yscale", "field": `${props.dimension}` },
+                    "height": { "signal": "bulletGroupHeight" },
+                    "width": { "signal": "width" }
+                }
+            },
+            "marks": []
+        }
+    ]
+
+    bulletMark[0].marks?.push(getBulletMarkRect(props));
```
spenjnb, line 96: This another suggestion maybe we could define an array of marks first and the spread them inside bulletMark[0].marks so that we don't need to do multiple '.push()' calls. It could help with readability and maintaining the code in case we need to add new marks later.

spenjnb: It could something like this:
const markElements = [
    getBulletMarkRect(props),
    getBulletMarkTarget(props),
    getBulletMarkLabel(props),
    getBulletMarkValueLabel(props)
];

bulletMark[0].marks = [...markElements];

src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+            "cornerRadiusBottomLeft": [
+                { "test": "datum.amount < 0", "value": 3 }
+            ],
+            "cornerRadiusTopRight": [
+                { "test": "datum.amount > 0", "value": 3 }
+            ],
+            "cornerRadiusBottomRight": [
+                { "test": "datum.amount > 0", "value": 3 }
+            ],
+            "fill": [{ "value": `${props.color}` }]
```
marshallpete, line 125:
```suggestion
            "fill": [{ "value": `${getColorValue(props.color, props.colorScheme)}` }]
```

marshallpete: This will make so that something like `gray-800` will work.

src/specBuilder/bullet/markUtils.ts
marshallpete, line 76:
```suggestion
export function getBulletMarks(props: BulletSpecProps): GroupMark {
```

src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+            "encode": {
+                "update": {
+                    "y": { "scale": "yscale", "field": `${props.dimension}` },
+                    "height": { "signal": "bulletGroupHeight" },
+                    "width": { "signal": "width" }
+                }
+            },
+            "marks": []
+        }
+    ]
```
marshallpete, line 94:
```suggestion
    const bulletMark: GroupMark = 
        {
            "name": "bulletGroup",
            "type": "group",
            "from": {
                "facet": { "data": "table", "name": "bulletGroups", "groupby": `${props.dimension}` }
            },
            "encode": {
                "update": {
                    "y": { "scale": "yscale", "field": `${props.dimension}` },
                    "height": { "signal": "bulletGroupHeight" },
                    "width": { "signal": "width" }
                }
            },
            "marks": []
        };
    
```

src/specBuilder/bullet/markUtils.ts
marshallpete: bulletMark doesn't need to be an array I don't think. It can just be a single GroupMark.

src/specBuilder/bullet/markUtils.ts
marshallpete: This looks correct to me. By providing `fields` the spec will find the min/max values in either field and set that to the scale min/max and since `zero: true` it will also make sure that 0 is in the scale domain.

src/specBuilder/bullet/markUtils.test.ts
```diff
@@ -0,0 +1,93 @@
+/*
```
marshallpete, line 1: Can you rename this file `bulletMarkUtils.ts`?
We like to have unique names for every file so that when you use CMD + P to find files, it's very clear what they are for.

src/specBuilder/bullet/bulletSpecBuilder.ts
```diff
@@ -49,143 +50,7 @@ export const addBullet = (
...
-      "description": `${props.name}`,
-      "encode": {
-        "enter": {
-          "x": {"value": 0},
-          "y": {"field": "index", "mult": 60, "offset": -44},
-          "width": {"scale": "xscale", "field": `${props.metric}`},
-          "height": {"value": 6},
-          "fill": {"value": `${props.color}`},
-          "cornerRadiusTopRight": {"value": 2},
-          "cornerRadiusBottomRight": {"value": 2}
```
marshallpete, line 75: Moving all of this is great

src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
+/*
```
marshallpete, line 1: Can you rename this file `bulletMarkUtils.ts`?
We like to have unique names for every file so that when you use CMD + P to find files, it's very clear what they are for.

src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+        { "name": "targetHeight", "update": "bulletThresholdHeight + 6" },
+        { "name": "bulletGroupHeight", "update": "bulletThresholdHeight + 24" },
+        { "name": "paddingRatio", "update": "gap / (gap + bulletGroupHeight)" },
+        {
+            "name": "height",
+            "update": "length(data('table')) * bulletGroupHeight + (length(data('table')) - 1) * gap"
+        }
+    ]
+
+    return bulletSignal;
```
marshallpete, line 53:
```suggestion
    const bulletSignals: Signal[] = [
        { "name": "gap", "value": 12 },
        { "name": "bulletHeight", "value": 8 },
        { "name": "bulletThresholdHeight", "update": "bulletHeight * 3" },
        { "name": "targetHeight", "update": "bulletThresholdHeight + 6" },
        { "name": "bulletGroupHeight", "update": "bulletThresholdHeight + 24" },
        { "name": "paddingRatio", "update": "gap / (gap + bulletGroupHeight)" },
        {
            "name": "height",
            "update": "length(data('table')) * bulletGroupHeight + (length(data('table')) - 1) * gap"
        }
    ]

    return bulletSignals;
```

src/specBuilder/bullet/markUtils.ts
```diff
@@ -0,0 +1,212 @@
...
+            "cornerRadiusBottomRight": [
+                { "test": "datum.amount > 0", "value": 3 }
+            ],
+            "fill": [{ "value": `${props.color}` }]
+            },
+            "update": {
+                "x": { "scale": "xscale", "value": 0 },
+                "x2": { "scale": "xscale", "field": `${props.metric}` },
+                "height": { "signal": "bulletHeight" },
+                "y": { "signal": "bulletGroupHeight - 3 - 2*bulletHeight" }
```
marshallpete, line 131:
```suggestion
                "y": { "signal": "bulletGroupHeight - 3 - 2 * bulletHeight" }
```

src/stories/components/Bullet/Bullet.story.tsx
```diff
@@ -24,18 +24,18 @@ export default {
...
+    height: 600,
 };
 
 // Basic Bullet chart story
 const BulletStory: StoryFn<BulletProps & { width?: number; height?: number }> = (args): ReactElement => {
     const { width, height, ...bulletProps } = args;
     const chartProps = useChartProps({ ...defaultChartProps, width: width ?? 350, height: height ?? 350 });
     return (
-        <Chart {...chartProps}>
+        <Chart {...chartProps} debug>
```
marshallpete, line 38:
```suggestion
        <Chart {...chartProps}>
```

src/specBuilder/bullet/markUtils.ts
rencherg: Color is already set using `getColorValue()` on line 42 in the `bulletSpecBuilder` file. Should we move the call to that function into the utils?

src/specBuilder/bullet/markUtils.ts
rencherg: I can do that but I think I would have to change `bulletSpecBuilder.ts` on line 52 to something like this:

`marks: [getBulletMarks(bulletProps)],`

It works as a standalone object in the vega playground but in the project the typing is requiring an array

PR #533, opened 2025-03-26
Scrum 28/flexible max scale

rencherg:
## Description

Adds functionality for fixed and flexible labeling

## How Has This Been Tested?

Relevant tests have been added

## Types of changes

## Checklist:


src/specBuilder/bullet/bulletMarkUtils.ts
```diff
@@ -97,6 +107,14 @@ export function getBulletData(props: BulletSpecProps): Data[] {
...
 	];
 
+	if (props.scaleType === 'flexible') {
+		bulletData[0].transform?.push({
+			type: 'formula',
+			expr: `${props.maxScaleValue}`,
+			as: 'flexibleScaleValue',
+		});
+	}
+
```
spenjnb, line 117: It could be good to add comment here to explain maxScaleValue. Like "Domain is at least `maxScaleValue` if data is smaller, but can grow if data is larger". `flexibleScaleValue` can have a comment that explains that this value is used to guarantee the domain includes at least `maxScaleValue` or if we don't want to add a comment maybe renaming it to `maxPossibleValue` could be more self-explanatory just an idea.

src/types/Chart.ts
```diff
@@ -243,6 +243,10 @@ export interface BulletProps extends MarkProps {
 	numberFormat?: NumberFormat;
 	/** Specifies if the labels should be in top of the bullet chart or to the side. Side labels are not supported in row mode. */
 	labelPosition?: 'side' | 'top';
+	/** Specifies if the scale should be normal, fixed, or flexible */
+	scaleType?: 'normal' | 'fixed' | 'flexible';
+	/** Maximum value for the scale */
+	maxScaleValue?: number;
```
spenjnb, line 249: When we set scale to `flexible` if the data is bigger than maxScaleValue the domain can grow beyond that. We could maybe expand this comment to explain that this max value can be excided, or maybe give it a different name like minMaxScaleValue, just to make clear that `flexible` ensures maxScaleValue but can go larger if data is bigger.

src/specBuilder/bullet/bulletMarkUtils.ts
```diff
@@ -17,6 +17,14 @@ import { getColorValue } from '../specUtils';
 export function getBulletScales(props: BulletSpecProps): Scale[] {
 	const groupScaleRangeSignal = props.direction === 'column' ? 'height' : 'width';
 	const xRange = props.direction === 'column' ? 'width' : [0, { signal: 'bulletGroupWidth' }];
+	let domainFields;
+	if (props.scaleType === 'flexible') {
+		domainFields = { data: 'table', fields: ['xPaddingForTarget', props.metric, 'flexibleScaleValue'] };
+	} else if (props.scaleType === 'fixed') {
+		domainFields = [0, `${props.maxScaleValue}`];
```
spenjnb, line 24: Do we want to handle negative data? The current implementation uses 0 for `fixed`, `normal` and `flexible`? Maybe we can just add one more test case for negative values.

rencherg: I changed to force positive values. A test was added

c-lamoureux (approved):

PR #534, opened 2025-03-31
Scrum 24/bullet track v2

rencherg:
## Types of changes

## Checklist:


src/specBuilder/bullet/bulletMarkUtils.test.ts
```diff
@@ -72,6 +73,18 @@ describe('getBulletMarks', () => {
 		const targetValueMark = marksGroup.marks?.find((mark) => mark.name?.includes('TargetValueLabel'));
 		expect(targetValueMark).toBeDefined();
 	});
+
+	test('Should include bullet track and not bullet threshold when props are configured correctly', () => {
```
c-lamoureux, line 77: Can we be a bit more specific than this. I'm not totally sure what the test intention is. Are we expecting that the track overrides the threshold if it's true?

src/types/Chart.ts
```diff
@@ -264,7 +264,11 @@ export interface BulletProps extends MarkProps {
 	 *
 	 * `fill` : The fill color to use for the threshold background.
 	 */
-	thresholds?: ThresholdBackground[];
+	thresholdValues?: ThresholdBackground[];
+	/** Color regions that sit behind the bullet bar */
+	threshold?: boolean,
```
c-lamoureux, line 269: Do we need this prop? We can probably assume that threshold === true if there are `thresholdValues`, or what we were previously calling `thresholds`. 

I think `thresholds` makes more sense since we're treating a `threshold` as a thing. Whereas `value` implies that it's a single value rather than an object. I'm guessing that was changed for clarity between `thresholds` and `threshold`. 

What do you think? Let me know if I'm missing context here.

src/specBuilder/bullet/bulletMarkUtils.ts
```diff
@@ -399,3 +402,42 @@ export function getBulletMarkThreshold(props: BulletSpecProps): Mark {
 	};
 	return bulletMarkThreshold;
 }
+
+export function getBulletMarkTrack(props: BulletSpecProps): Mark {
+
+	const trackColor = getColorValue('gray-200', props.colorScheme);
+	const trackWidth = props.direction === 'column' ? 'width' : 'bulletGroupWidth';
+    
+	const bulletTrackMark: Mark = {
```
c-lamoureux, line 411: We should call this `bulletMarkTrack` or change the function to `getBulletTrackMark`. We could also just do `getBulletTrack` since this is the only context where we're planning on using `track`.

src/specBuilder/bullet/bulletMarkUtils.test.ts
```diff
@@ -72,6 +73,18 @@ describe('getBulletMarks', () => {
...
+		const props = { ...samplePropsColumn, threshold: false, track: true };
+		const marksGroup = getBulletMarks(props);
+		console.log(marksGroup);
```
c-lamoureux, line 80:
```suggestion
```

c-lamoureux (changes requested):
Can we add a new story for this?

src/specBuilder/bullet/bulletMarkUtils.ts
rencherg: Changed to get bulletTrack

src/specBuilder/bullet/bulletMarkUtils.ts
```diff
@@ -141,9 +142,9 @@ export function getBulletMarks(props: BulletSpecProps): GroupMark {
 		marks: [],
 	};
 
-	const thresholdValues = props.thresholds;
+	const thresholdValues = props.thresholdValues;
 
-	if (thresholdValues) {
+	if (props.threshold && thresholdValues) {
```
spenjnb, line 147: If we set both `threshold: true` and `track: true` it only renders thresholds. Is this intended or should we allow both to be drawn together?

src/types/Chart.ts
```diff
@@ -264,7 +264,11 @@ export interface BulletProps extends MarkProps {
 	 *
 	 * `fill` : The fill color to use for the threshold background.
 	 */
-	thresholds?: ThresholdBackground[];
+	thresholdValues?: ThresholdBackground[];
```
spenjnb, line 267: Now that we renamed `thresholds` to `thresholdValues`, some places (like `Thresholds.args`) still use both.
```
threshold: true,
thresholdValues: basicThresholdsData
```
Should we update the rest of the variable names?

src/types/Chart.ts
rencherg: Uploading Screen Recording 2025-04-01 at 6.19.40 PM.mov…

I added the prop because for some reason removing the data from the thresholds field would not remove the threshold from the visualization. The only way I could remove it was by adding an empty list or empty object, however it doesn't make sense to ask users to do this. Adding the prop was the best remaining way in my opinion.

src/specBuilder/bullet/bulletMarkUtils.ts
rencherg: Yes, this behavior is intended. If both are true then it shows the threshold instead of the track

src/types/Chart.ts
c-lamoureux: It might be storybook not updating if no value is provided to the control. What happens if we create a track specific story and just leave thresholds empty? Or is thresholds defaulting to an empty array?

src/specBuilder/bullet/bulletMarkUtils.ts
```diff
@@ -152,6 +153,8 @@ export function getBulletMarks(props: BulletSpecProps): GroupMark {
 			},
 		];
 		bulletMark.marks?.push(getBulletMarkThreshold(props));
+	} else if(props.track) {
```
c-lamoureux, line 156: It looks like this wasn't picked up by the linter.

c-lamoureux:
This is pretty much good. Let's keep the prop name as `thresholds` and double check that code is formatted correctly. I saw a case where it wasn't and left a comment. Once that's done this will be good to go.

PR #536, opened 2025-04-01
Scrum 35/refactor/bullet scales and signals produce

spenjnb:
## Description

This PR refactors the Bullet chart’s getBulletScales and getBulletSignals functions to use the produce pattern from Immer. The refactored versions are now named addScales and addSignals and follow the same structure as other chart components (e.g., Bar, Donut).

## Related Issue

General refactoring.

## Motivation and Context

- Aligns Bullet chart logic with the rest of the chart implementations.
- Enables a standardized, immutable update pattern via produce.

## How Has This Been Tested?

- Updated the Bullet unit tests to pass the additional parameter ([]) required by the produce functions.
- Verified that Bullet renders correctly in Storybook and behaves as expected with both default and custom props.
- Ran the full test suite to confirm no regressions across other parts of the codebase.

- [x] Refactor (non-breaking change improving code structure)

## Checklist:

### DIFF
--- a/src/specBuilder/bullet/bulletMarkUtils.ts
+++ b/src/specBuilder/bullet/bulletMarkUtils.ts
@@ -15,7 +15,7 @@ import { Axis, Data, GroupMark, Mark, Scale, Signal } from 'vega';
 import { BulletSpecProps } from '../../types';
 import { getColorValue } from '../specUtils';
 
-export const addBulletScales = produce<Scale[], [BulletSpecProps]>((scales, props) => {
+export const addScales = produce<Scale[], [BulletSpecProps]>((scales, props) => {
 	const groupScaleRangeSignal = props.direction === 'column' ? 'height' : 'width';
 	const xRange = props.direction === 'column' ? 'width' : [0, { signal: 'bulletGroupWidth' }];
 	let domainFields;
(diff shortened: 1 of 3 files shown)


c-lamoureux:
This looks great! Looks like there's a conflict to resolve before we can merge.

src/specBuilder/bullet/bulletMarkUtils.ts
```diff
@@ -42,53 +43,52 @@ export function getBulletScales(props: BulletSpecProps): Scale[] {
...
+	signals.push({ name: 'bulletThresholdHeight', update: 'bulletHeight * 3' });
+	signals.push({ name: 'targetHeight', update: 'bulletThresholdHeight + 6' });
 
 	if (props.showTargetValue && props.showTarget) {
-		bulletSignals.push({ name: 'targetValueLabelHeight', update: '20' });
+		signals.push({ name: 'targetValueLabelHeight', update: '20' });
 	}
 
-	bulletSignals.push({
+	signals.push({
```
rencherg, line 60: Interesting choice for removing the word bullet. Do we want to do that for the rest of the util file?

rencherg (approved):

src/specBuilder/bullet/bulletMarkUtils.ts
spenjnb: Good comment, so I removed bullet from the variable name here because we’re now using `produce` to update the `signals` array directly, so we’re not creating a separate bulletSignals array anymore. That said, I'm happy to change the naming to match the rest of the file either way, what do you think?

rencherg: Thanks for the explanation. I guess if we switch any other methods to produce we should remove bullet but if not we should leave it as is

c-lamoureux (approved):
Nice!

PR #541, opened 2025-04-07
Scrum 32/bar label color change

spenjnb:
## Description

This PR implements the bullet chart threshold-coloring feature, along with the necessary logic to handle partial or missing thresholds. Specifically:

Threshold-Based Bar Color

- We generate a nested ternary expression (or an array of conditional rules, depending on the approach) that assigns a fill color based on the metric value’s range.
- If thresholds are missing or incomplete, we fall back to the default color.

Metric Label Color Matching

- If the computed threshold color is not the default, the metric’s label also adopts the dynamic color.
- If it is the default, we revert the label color to a standard neutral (e.g. gray/black) for readability.

## Related Issue

SCRUM-32

## Motivation and Context

To adjust color based on user-defined threshold ranges, making it easier to visually communicate performance or progress.

## How Has This Been Tested?

- Added tests for the expression helper to ensure correct generation of color expressions for 0, 1, 2, or 3 thresholds.
- Updated the bullet chart tests to confirm that the bar and label coloring match the expected threshold color logic under different chart data scenarios.
- Verified that removing thresholds while retaining a higher or lower metric does not result in parse errors.
- Confirmed that the label color reverts to a default if no matching threshold is found.

## Types of changes

## Checklist:


src/specBuilder/bullet/bulletDataUtils.ts
```diff
@@ -56,5 +56,58 @@ export const getBulletTransforms = (props: BulletSpecProps): FormulaTransform[]
...
+	});
+
+	const exprParts: string[] = [];
+
+	// For values below the first threshold's lower bound, use the default color.
+	exprParts.push(
+		`(datum.${metricField} < ${
+			sorted[0].thresholdMin !== undefined ? sorted[0].thresholdMin : -1e12
+		}) ? '${defaultColor}' : `
+	);
```
marshallpete, line 99: Instead of using `-1e12` does it work to use `-Infinity`?

marshallpete:
Thanks for all the stories and tests

src/specBuilder/bullet/bulletDataUtils.ts
spenjnb: Yes, so I initially tried using Infinity, but Vega did not recognize the expression. I was getting this error: "Uncaught (in promise) Error: Unrecognized signal name: 'Infinity'".

rencherg (approved):

PR #571, opened 2025-04-28
Bar dual metric axis

c-lamoureux:
## Description

a `dualMetricAxis` prop for bar. This adds metric based colors to the axis and opacity to the irrelevant metric when hovering a bar for the opposing axis. 

We may add a more controlled approach with direct mapping of metrics to axes in the future. 

## Motivation and Context

This is a simple solution for consumers who would like to add a single prop value to a bar with 2 metric axes in order to get clean dual axis scaling. (vertical bar: left | right, horizontal bar: top | bottom).

This is implemented separately from the `Combo` chart. We felt that having a `Combo` for multiple charts of the same chart type would be unnecessarily complex both to implement and to consume as opposed to this single prop. We may unify some implementation details for this and `Combo` if we find it necessary in the future.

## How Has This Been Tested?

Thorough testing for spec builder and rendering of stories.

Screenshot 2025-04-28 at 13 59 50
Screenshot 2025-04-28 at 14 00 00
Screenshot 2025-04-28 at 14 00 10

## Types of changes

## Checklist:


packages/vega-spec-builder/src/axis/axisSpecBuilder.ts
```diff
@@ -9,16 +9,24 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
```
marshallpete, line 11: Would be good to split this file up in a follow up PR.

packages/vega-spec-builder/src/bar/barSpecBuilder.ts
```diff
@@ -166,6 +192,10 @@ export const addData = produce<Data[], [BarSpecOptions]>((data, options) => {
 	if (type === 'dodged' || isDodgedAndStacked(options)) {
 		data[index].transform?.push(getDodgeGroupTransform(options));
 	}
+
+	if (isDualMetricAxis(options)) {
+		addDualMetricAxisData(data, options);
+	}
```
marshallpete, line 198: Can we move the if statement into the `addDualMetricAxisData()` function instead of here? Similar to what we do for `addTendlineData`, `addTooltipData`, and `addPopoverData`

packages/vega-spec-builder/src/bar/barSpecBuilder.ts
```diff
@@ -224,13 +254,47 @@ export const getDodgeGroupTransform = ({ color, lineType, name, opacity, type }:
 	};
 };
 
+export const addDualMetricAxisData = (data: Data[], options: BarSpecOptions) => {
+	const baseScaleName = getBaseScaleName(options);
```
marshallpete, line 258:
```suggestion
	if (!isDualMetricAxis(options)) return;
	const baseScaleName = getBaseScaleName(options);
```

packages/vega-spec-builder/src/types/specUtil.types.ts
```diff
@@ -34,11 +34,14 @@ export interface ScSpec extends Spec {
...
 	 * A preferred approach would be to add the needed information to usermeta { orientation: 'vertical' }.
 	 */
-	usermeta?: {
-		orientation?: Orientation;
-	};
+	usermeta: UserMeta;
 }
 
+export type UserMeta = {
+	orientation?: Orientation;
```
marshallpete, line 41: Can we rename this to be a bit more specific? Even something like `chartOrientation`. We could easily support orientation on things like axis labels, legends etc.

PR #627, opened 2025-08-20
Feat/axis hover tooltip

stevenshow:
## Description

Added a tooltip to the axis. Extended the tooltip delay from legend items to axis items. Updated the tooltip delay constant to be more generic.

## Motivation and Context

As we move towards the axis containing more lengthy strings, we are adding the tooltip functionality there so that the user can still access that data.

## How Has This Been Tested?

I updated the spec tests the include the two new props `Interactive` as well as `tooltip`

image

## Types of changes

## Checklist:

### DIFF
--- a/packages/constants/constants.ts
+++ b/packages/constants/constants.ts
@@ -123,7 +123,7 @@ export const DEFAULT_VENN_LABEL = 'label';
 export const HIGHLIGHT_CONTRAST_RATIO = 5;
 
 // legend tooltips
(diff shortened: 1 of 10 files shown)


c-lamoureux:
Looks solid! Legend tooltips only appear when there are descriptions on the legend. We don't need axis descriptions, but we should add a prop like `hasTooltip` to the axis so this isn't enabled by default. We may rework how we control it in v2.

c-lamoureux (approved):

PR #640, opened 2025-10-01
refactor: series highlighting refactor

marshallpete:
refactor series highlighting.

HIGHLIGHTED_SERIES is now for controlled methods only.

Everything else is name spaced.

Added a CONTROLLED_HIGHLIGHT_TABLE

### DIFF
--- a/packages/vega-spec-builder/src/line/lineMarkUtils.ts
+++ b/packages/vega-spec-builder/src/line/lineMarkUtils.ts
@@ -12,13 +12,14 @@
 import { LineMark, Mark, NumericValueRef, ProductionRule, RuleMark } from 'vega';
 
 import {
+  CONTROLLED_HIGHLIGHTED_TABLE,
   DEFAULT_INTERACTION_MODE,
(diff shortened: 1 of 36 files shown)


packages/vega-spec-builder/src/line/lineMarkUtils.ts
```diff
@@ -104,24 +105,29 @@ export const getLineOpacity = ({
...
+      });
+    }
   }
 
-  // add a rule that will lower the opacity of the line if there is a hovered series, but this line is not the one hovered
   strokeOpacityRules.push(
     {
-      test: `isValid(${HIGHLIGHTED_SERIES}) && ${HIGHLIGHTED_SERIES} !== datum.${SERIES_ID}`,
-      value: FADE_FACTOR,
+      test: `length(data('${CONTROLLED_HIGHLIGHTED_TABLE}'))`,
```
c-lamoureux, line 124: Love that we can start to do this instead of so many complex chains of `isValid` checks!

c-lamoureux (approved):
Amazing work!

PR #805, opened 2026-05-18
S2 Line: alternate segment style 

madelineluke:
## Description

adding alternate segment to line, adds the ability to switch between solid and dotted/dashed in the same line

New props on <Line> (S2)
  
  - alternateSegmentKey — a data field key whose truthy value marks a point as part of an alternate segment
  - alternateSegmentLineType — the line type used for alternate segments (defaults to 'dotted')
  - alternateSegmentLabel — text to append to hover labels for alternate-segment points (stored but not yet wired to tooltip, will be usable after the LinePointAnnotation is implemented)

## Motivation and Context

Adds the ability to switch between solid and dotted/dashed lineType in the same line 

## How Has This Been Tested?

Unit testing and storybook visual testing. 

Screenshot 2026-05-18 at 8 43 43 AM

## Types of changes

## Checklist:

### DIFF
--- a/packages/vega-spec-builder-s2/src/line/lineMarkUtils.ts
+++ b/packages/vega-spec-builder-s2/src/line/lineMarkUtils.ts
@@ -9,7 +9,7 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-import { LineMark, Mark, NumericValueRef, ProductionRule, RuleMark } from 'vega';
+import { ArrayValueRef, LineMark, Mark, NumericValueRef, ProductionRule, RuleMark } from 'vega';
 
 import {
   COLOR_SCALE,
@@ -38,6 +38,7 @@ import {
   getXProductionRule,
   hasPopover,
 } from '../marks/markUtils';
+import { getStrokeDashFromLineType } from '../specUtils';
 import { getDualAxisScaleNames } from '../scale/scaleUtils';
 import { ScaleType } from '../types';
 import {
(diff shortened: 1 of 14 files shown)


packages/vega-spec-builder-s2/src/line/lineMarkUtils.ts
```diff
@@ -127,6 +128,25 @@ const getGradientOpacity = (opacity: LineMarkOptions['opacity']): { value: numbe
...
+ * Falls back to the standard scale/field lookup for data-driven lineType facets.
+ */
+export const getAlternateSegmentStrokeDash = (
+  name: string,
+  lineType: LineMarkOptions['lineType'],
+  alternateSegmentLineType: LineMarkOptions['alternateSegmentLineType']
+): ArrayValueRef => {
+  if (typeof lineType !== 'string' && !Array.isArray(lineType)) {
+    const baseDash = JSON.stringify(getStrokeDashFromLineType(lineType.value));
+    const altDash = JSON.stringify(getStrokeDashFromLineType(alternateSegmentLineType ?? 'dotted'));
```
c-lamoureux: Can we set the default alternate line type higher up in the chain where we set other prop defaults? Worried it will be harder to understand behavior at a glance by putting it here.

packages/vega-spec-builder-s2/src/line/lineSpecBuilder.ts
```diff
@@ -270,6 +286,53 @@ const getMetricKeys = (lineOptions: LineSpecOptions) => {
   return metricKeys;
 };
 
+export const getAlternateSegmentData = (name: string, dimSortField: string): Data[] => [
```
c-lamoureux, line 289: We're sorting this data by ascending, does that risk us effecting the order of data as it's passed in? For instance, if the consumer passes data in an unconventional order or descending in value, would this change that?

I'm investigating now as I'm not certain but figured you'd have more context.

packages/vega-spec-builder-s2/src/line/lineMarkUtils.ts
madelineluke: yes!

packages/vega-spec-builder-s2/src/line/lineSpecBuilder.ts
madelineluke: no it shouldn't. It is sorting the way that the window processes the rows, but it is not re-ordering the rows. I will double check this

c-lamoureux:
It looks like this doesn't handle key based line types:
```
const AlternateSegmentDataDrivenLineType = bindWithProps(AlternateSegmentStory);
AlternateSegmentDataDrivenLineType.args = {
  color: 'series',
  name: 'line0',
  lineType: 'series',
  alternateSegmentKey: 'isEstimated',
  alternateSegmentLineType: 'dotted',
};
```

packages/vega-spec-builder-s2/src/line/lineSpecBuilder.ts
c-lamoureux: Yeah makes sense to me after reviewing further. Thanks!

c-lamoureux (approved):

PR #815, opened 2026-06-03
feat: S/M/L font scaling

MeepoBleepo:
## Description

Added auto scaling for Direct Labels so that their font size adjusts between S/M/L depending on chart size.

## Motivation and Context

The S2 design spec requires that direct label font size and weight scale down for medium and small chart contexts — a direct label appropriate for a large full-page chart is visually heavy on a sparkline or trellis panel.

## How Has This Been Tested?

Unit tests verify that:
- Both marks emit { signal: 'rscChartSizeFontSize' } and { signal: 'rscChartSizeFontWeight' } by default
- When fonSize/fontWeight are explicitly set on the label options, the marks emit use that value instead
- When fontSize/fontWeight is not explicitly set, they are passed as undefined

Also added new storybook story "Direct Label Size Scaling" for visually testing behavior.

image
image
image

## Types of changes

## Checklist:

### DIFF
--- a/packages/constants/constants.ts
+++ b/packages/constants/constants.ts
@@ -194,11 +194,11 @@ export const BACKGROUND_COLOR = 'chartBackgroundColor';
 export const DIRECT_LABEL_FONT_WEIGHT = 700 as const;
 export const DIRECT_LABEL_BACKGROUND_STROKE_WIDTH = 4;
 
-/** Direct label font size consants */
+/** Direct label font size constants */
 export const DIRECT_LABEL_FONT_SIZE_S = 14;
 export const DIRECT_LABEL_FONT_SIZE_M = 16;
 export const DIRECT_LABEL_FONT_SIZE_L = 18;
-/** Direct label font weight consants */
+/** Direct label font weight constants */
 export const DIRECT_LABEL_FONT_WEIGHT_S = 700 as const;
 export const DIRECT_LABEL_FONT_WEIGHT_M = 700 as const;
 export const DIRECT_LABEL_FONT_WEIGHT_L = 700 as const;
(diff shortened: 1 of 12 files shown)


packages/constants/constants.ts
```diff
@@ -157,6 +161,15 @@ export const BACKGROUND_COLOR = 'chartBackgroundColor';
 export const DIRECT_LABEL_FONT_WEIGHT = 700 as const;
 export const DIRECT_LABEL_BACKGROUND_STROKE_WIDTH = 4;
 
+/** Direct label font size constants */
+export const DIRECT_LABEL_FONT_SIZE_S = 14;
+export const DIRECT_LABEL_FONT_SIZE_M = 16;
+export const DIRECT_LABEL_FONT_SIZE_L = 18;
+/** Direct label font weight constants */
+export const DIRECT_LABEL_FONT_WEIGHT_S = 700 as const;
```
madelineluke: if the font weight stays the same across all sizes, we can just use the DIRECT_LABEL_FONT_WEIGHT constant!

packages/vega-spec-builder-s2/src/lineDirectLabel/lineDirectLabelUtils.ts
```diff
@@ -181,7 +185,10 @@ export const getLineDirectLabelMarks = (
 				...baseEnter,
 				fill: getColorProductionRule(labelOptions.color, labelOptions.colorScheme),
 			},
-			update: { fontWeight: { value: DIRECT_LABEL_FONT_WEIGHT }, opacity: opacityRules },
+			update: {
```
madelineluke, line 205: it looks like we used to set `{ fontWeight: { value: DIRECT_LABEL_FONT_WEIGHT }`, we still want to keep this so it stays at 700, we just don't need the signal with the changing S/Ml/L values since it stays the same

madelineluke (dismissed):

madelineluke (approved):

PR #819, opened 2026-06-08
fix: legend columns account for actual label  size rather than just using labelLimit

c-lamoureux:
## Description

Calculate the actual label text size to determine columns, rather than just relying on the labelLimit. Columns are calculated using the largest actual label.

## Motivation and Context

This makes layouts where the labelLimit is higher, but the actual label sizes are small. Prior to this change, this type of layout would create a small number of columns even though there is plenty of space to spread the legend items out. This change will make legend layouts more natural. 

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.


madelineluke:
Looks nice, do we want to add stories to the S2 storybook as well?

madelineluke (approved):

PR #825, opened 2026-06-15
bug: s2 chart title wrap 

madelineluke:
Implementing an expression function that is used in the title spec signal to wrap the title text when it is long or there is not enough space. 

`wrapTitleText()`: calculates the max number of characters allowed per line based on the width of the chart and an estimated size of the text. separates each word out into an array, adds words to that line until the next word will exceed the limit, and then makes a new array entry. 

## Motivation and Context

Solves when a chart title is too long to fit on a single line within the chart's container width, the current S2 implementation shrinks the chart horizontally to accommodate the overflowing title text rather than wrapping the title onto multiple lines. This causes the data visualization area to contract unexpectedly and is visually broken for long or dynamic title strings.

## How Has This Been Tested?

Unit testing the new function to wrap test, visible testing in storybook. 

Screenshot 2026-06-15 at 4 26 52 PM

## Types of changes

## Checklist:


packages/vega-spec-builder-s2/src/specUtils.ts
```diff
@@ -247,6 +223,31 @@ export const initializeSpec = (spec: Spec | null = {}, chartOptions: Partial<Cha
   return { ...baseSpec, ...(spec || {}) };
 };
 
+// Average character width ratio for adobe-clean at the title font size
+const AVG_CHAR_WIDTH_RATIO = 0.5;
```
c-lamoureux: This may not be dynamic enough. 
Title accepts `fontWeight`. We might be able to estimate size using the `fontWeight` value, but we have an existing expression function that draws the actual text to get the exact size. 

If we could either use that or make a title specific function (if necessary), we wouldn't have to worry about future props effecting this size either.
https://github.com/adobe/react-spectrum-charts/blob/b278a8161d42f2d961eb3d521eb55b6acd257228/packages/vega-spec-builder-s2/src/expressionFunctions/expressionFunctions.ts#L181-L188

madelineluke: yeah, this was my biggest concern with this solution. Let me try using this existing function!

PR #828, opened 2026-07-06
s2 line hover labels

madelineluke:
## Description

Add hover labels to s2 line chart points, including dimension hover. The hover labels also have a cascading formula in place that offset the labels to avoid collision. 

## Motivation and Context

Provide point data info on line charts. 

## How Has This Been Tested?

Unit tests and storybook 

Screenshot 2026-07-06 at 4 40 52 PM

## Types of changes

## Checklist:

### DIFF
--- a/packages/constants/constants.ts
+++ b/packages/constants/constants.ts
@@ -77,6 +77,17 @@ export const CHART_SIZE_POINT_SIZES = {
   L: (BASE_POINT_DIAMETER * CHART_SIZE_SCALE_RATIOS.L) ** 2,
 } as const;
 
+/** Vega signal name for the chart-size-derived label gap, driven by a reactive expression. */
+export const CHART_SIZE_LABEL_GAP = 'rscChartSizeLabelGap';
+
+/** Minimum vertical gap (px) between co-located labels per chart size tier. Base M = 12px. */
+const BASE_LABEL_GAP = 12;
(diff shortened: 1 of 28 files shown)


packages/constants/constants.ts
```diff
@@ -77,6 +77,17 @@ export const CHART_SIZE_POINT_SIZES = {
   L: (BASE_POINT_DIAMETER * CHART_SIZE_SCALE_RATIOS.L) ** 2,
 } as const;
 
+/** Vega signal name for the chart-size-derived label gap, driven by a reactive expression. */
+export const CHART_SIZE_LABEL_GAP = 'rscChartSizeLabelGap';
+
+/** Minimum vertical gap (px) between co-located labels per chart size tier. Base M = 12px. */
+const BASE_LABEL_GAP = 12;
+export const CHART_SIZE_LABEL_GAPS = {
```
c-lamoureux, line 85: Did we make a prior decision to make label gaps scale with chart size?

madelineluke: yeah I decided to scale it because with the large size they were overlapping a little too much, I think I showed you when we had that call a few weeks ago. If it looks off though let me know

c-lamoureux: It looks good. I just forgot over the break. We set the constants up so it's easy to tweak later. Good stuff!

c-lamoureux (approved):

PR #833, opened 2026-07-09
feat: hover animations initial setup

MeepoBleepo:
## Description

Lays the groundwork for animated hover/highlight emphasis for marks. New file hoverAnimationUtils.ts defines functions for supporting this animation system.

## Motivation and Context

This is the core of the hover animation system, defining new data sources and signals which marks will need to utilize in order to begin animating their hover behavior.

## How Has This Been Tested?

**Unit testing:**
* hoverAnimationUtils.test.ts added for testing hoverAnimationUtils.ts functions  

**Storybook:**
* LineHoverAnimation.story.ts added 7 new stories for testing the various types of hover interactions as well as a stress-testing story for performance testing. These stories won't yet show any changes since the animation system is not yet hooked up to line marks.

## Types of changes

## Checklist:

### DIFF
--- a/packages/constants/constants.ts
+++ b/packages/constants/constants.ts
@@ -118,6 +118,8 @@ export const DEFAULT_LEGEND_LABEL_LIMIT = 184;
 export const TABLE = 'table';
 export const FILTERED_TABLE = 'filteredTable';
 export const CONTROLLED_HIGHLIGHTED_TABLE = 'controlledHighlightedTable';
+/** Single-row data source recording timestamp of the most recent hover target change */
+export const HOVER_ANIM_LAST_CHANGE_DATA = 'hoverAnimLastChangeData'; 
 
 // vega data field names
 export const DIMENSION_FIELD = 'rscDimensionField';
@@ -142,6 +144,8 @@ export const FIRST_RSC_SERIES_ID = 'firstRscSeriesId'; // first series for dual 
 export const LAST_RSC_SERIES_ID = 'lastRscSeriesId'; // last series for dual y-axis
 export const HOVER_TIMER = 'hoverTimer'; // hover animation timer signal
 export const HOVER_TARGETS = 'hoverTargets'; // hover animation target values
+export const HOVER_ANIMATING = 'hoverAnimating'; // hover animation state signal
+export const HOVER_ACTIVE_TIMER = 'hoverActiveTimer'; // animation timer to run only when hoverAnimating is true
 
 // scale names
 export const COLOR_SCALE = 'color';
@@ -168,8 +172,8 @@ export const LINEAR_PADDING = 0;
 export const TRELLIS_PADDING = 0.2;
 
 // hover animation constants
-/** Timer signal update interval in ms. Caps timer signal update at ~60fps. */
-export const ANIMATION_THROTTLE = 16;
+/** Timer signal update interval in ms. Caps timer signal update at ~30fps. */
+export const ANIMATION_THROTTLE = 33;
 /** Time in ms it takes to animate between hover states (hovered -> unhovered etc.) */
 export const ANIMATION_HOVER_SPEED = 100;
 /**
--- a/packages/constants/constants.ts
+++ b/packages/constants/constants.ts
@@ -146,6 +146,7 @@ export const HOVER_TIMER = 'hoverTimer'; // hover animation timer signal
 export const HOVER_TARGETS = 'hoverTargets'; // hover animation target values
 export const HOVER_ANIMATING = 'hoverAnimating'; // hover animation state signal
 export const HOVER_ACTIVE_TIMER = 'hoverActiveTimer'; // animation timer to run only when hoverAnimating is true
+export const HOVER_IDLE_TICKS = 'hoverIdleTicks'; // consecutive ticks since hoverAnimating went false
 
 // scale names
 export const COLOR_SCALE = 'color';
--- a/packages/constants/constants.ts
+++ b/packages/constants/constants.ts
@@ -172,6 +172,13 @@ export const TRELLIS_PADDING = 0.2;
 export const ANIMATION_THROTTLE = 16;
 /** Time in ms it takes to animate between hover states (hovered -> unhovered etc.) */
 export const ANIMATION_HOVER_SPEED = 100;
+/**
+ * The resting hover-animation target when nothing is hovered. The fraction is an "emphasis level":
+ * 0 = deemphasized (something else hovered), this = neutral (nothing hovered), 1 = emphasized (this hovered).
+ * Consumers map the lower half [0, neutral] for deemphasis (e.g. opacity) and the upper half [neutral, 1]
+ * for emphasis (e.g. stroke width).
+ */
+export const HOVER_NEUTRAL_TARGET = 0.5;
 
 // donut constants
 /** Calculation for donut radius, subtract 2 pixels to make room for the selection ring */
--- a/packages/constants/constants.ts
+++ b/packages/constants/constants.ts
@@ -140,6 +140,8 @@ export const SELECTED_SERIES = 'selectedSeries'; // series
 export const SELECTED_GROUP = 'selectedGroup'; // data point
 export const FIRST_RSC_SERIES_ID = 'firstRscSeriesId'; // first series for dual y-axis
 export const LAST_RSC_SERIES_ID = 'lastRscSeriesId'; // last series for dual y-axis
+export const HOVER_TIMER = 'hoverTimer'; // hover animation timer signal
+export const HOVER_TARGETS = 'hoverTargets'; // hover animation target values
 
 // scale names
 export const COLOR_SCALE = 'color';
@@ -164,6 +166,12 @@ export const DISCRETE_PADDING = 0.5;
 export const PADDING_RATIO = 0.4;
 export const LINEAR_PADDING = 0;
 export const TRELLIS_PADDING = 0.2;
+
+// hover animation constants
+/** Timer signal update interval in ms. Caps timer signal update at ~60fps. */
+export const ANIMATION_THROTTLE = 16;
+/** Time in ms it takes to animate between hover states (hovered -> unhovered etc.) */
+export const ANIMATION_HOVER_SPEED = 100;
 
 // donut constants
 /** Calculation for donut radius, subtract 2 pixels to make room for the selection ring */
--- a/packages/react-spectrum-charts-s2/src/stories/Line/Features/HoverAnimation/LineHoverAnimation.story.tsx
+++ b/packages/react-spectrum-charts-s2/src/stories/Line/Features/HoverAnimation/LineHoverAnimation.story.tsx
@@ -22,7 +22,7 @@ import { Chart } from '../../../../Chart';
 import { Axis, ChartInspect, ChartPopover, Legend, Line } from '../../../../components';
 import useChartProps from '../../../../hooks/useChartProps';
 import { workspaceTrendsData } from '../../../../stories/data/data';
-import { formatTimestamp, generateLargeData } from '../../../../stories/storyUtils';
+import { formatTimestamp } from '../../../../stories/storyUtils';
 import { bindWithProps } from '../../../../test-utils';
 import { ChartProps } from '../../../../types';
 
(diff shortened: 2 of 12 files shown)


packages/react-spectrum-charts-s2/src/stories/Line/Features/HoverAnimation/LineHoverAnimation.story.tsx
```diff
@@ -0,0 +1,208 @@
...
+// │ PopoverSelection    │ <ChartPopover> — click a point to select                 │ popoverMatch                                  │
+// ├─────────────────────┼──────────────────────────────────────────────────────────┼───────────────────────────────────────────────┤
+// │ ControlledHighlight │ highlightedSeries chart prop                             │ controlledSeriesMatch                         │
+// ├─────────────────────┼──────────────────────────────────────────────────────────┼───────────────────────────────────────────────┤
+// │ OnClick             │ onClick handler makes the line interactive               │ hoveredMatch (interactive-via-click)          │
+// └─────────────────────┴──────────────────────────────────────────────────────────┴───────────────────────────────────────────────┘
+
+/**
+ * Showcases the line hover-animation system across every interaction that can emphasize a series.
+ * In each story the emphasized series smoothly grows its stroke width while the others fade their
```
c-lamoureux, line 47: This says we're smoothly increasing the stroke width, is that something we changed as well?

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
```diff
@@ -0,0 +1,160 @@
+/*
```
c-lamoureux, line 1: For some reason this file doesn't seem to be formatted properly. Missing spaces. Can we do a lint run on it?

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+    name: string;
+    groupby: string[];
+    rules: HoverMatchRule[];
+    source?: string; // defaults to FILTERED_TABLE
+}
+
+/**
+ * Constructs the data source for checking the hover interaction rules, storing the hover state as target values for each hoverable item.
+ * HoverMatchRules are defined per mark and injected into this data source.
+ * @param param0
```
c-lamoureux: We should make these parameters in the function comments accurate. AI is not great at writing these in a way that communicates intent so it may be necessary to hand-write.

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+ * @returns 
+ */
+export const getHoverTargetData = ({
+    name, groupby, rules, source = FILTERED_TABLE,
+}: HoverTargetDataOptions): SourceData => {
+    const transforms: (AggregateTransform | FormulaTransform)[] = [
+        { type: 'aggregate', groupby: groupby },
+        ...rules.map<FormulaTransform>((r) => ({ type: 'formula', as: r.as, expr: r.expr })),
+    ];
+    const targetExpr = rules.map((r) => `isValid(datum.${r.as}) ? datum.${r.as}`).join(' : ') + ` : ${HOVER_NEUTRAL_TARGET}`;
```
c-lamoureux: Can we add a comment showing an example of what the completed rule with datum.${r.as} would look like in practice? This is abstract by necessity, so it would be nice to have a clear example visible.

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+export const getHoverAnimStateData = ({
+    name, keys, keyField = SERIES_ID,
+}: HoverAnimStateOptions): ValuesData => ({
+    name: `${name}_hoverAnimStateData`,
+    values: keys.map((id) => ({ [keyField]: id, startTime: 0, startValue: 1, target: 1 })),
+    on: keys.map((_, i) => getOnTriggerEntry(name, i)),
+});
+
+/**
+ * Helper function to get the on trigger entry for each hoverable item.
```
c-lamoureux: This would be a good place to provide some brief detail on the application intent. What does the trigger entry ultimately do when we hover a line? Doesn't have to be granular, we can look at the code for it, but a brief explanation helps us connect what we see in the UI while reading through the code.

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * Returns a Vega expression string that evaluates to the current animation fraction (0..0.5..1)
+ * for `datum`, looked up by `keyField`. Marks compose this into whatever visual property they
+ * want to animate (opacity, strokeWidth, radius, ...). The engine only exposes the raw fraction.
+ * @param name mark name
+ * @param keyField identity field used to match the animation row (defaults to SERIES_ID)
+ */
+export const getHoverFractionSignal = (name: string, keyField: string = SERIES_ID): string => {
+    const fractionData = `data('${name}_hoverFractionData')`;
+    const lookup = `indexof(pluck(${fractionData}, '${keyField}'), datum.${keyField})`;
```
c-lamoureux: We're using the key to look up the index of the fraction data and then returning `fractionData[index]` below. Is there a reason we can't use the key for direct access within the signal itself?

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+export const getDeemphasisRamp = (fractionExpr: string): string =>
+    `clamp(${fractionExpr} / ${HOVER_NEUTRAL_TARGET}, 0, 1)`;
+
+/**
+ * Reacts only while a series is being emphasized; ignores de-emphasis. Concretely:
+ *   deemphasized (0) -> 0   ·   neutral -> 0   ·   emphasized (1) -> 1
+ * (Returns a flat 0 from neutral downward, and ramps to 1 as the series is pushed above neutral.)
+ */
+export const getEmphasisRamp = (fractionExpr: string): string =>
+    `clamp((${fractionExpr} - ${HOVER_NEUTRAL_TARGET}) / (1 - ${HOVER_NEUTRAL_TARGET}), 0, 1)`;
```
c-lamoureux: Can we provide some additional detail on the formula itself and what it's doing? If it's based on a specific formula we should reference that here as well.

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
```diff
@@ -0,0 +1,160 @@
...
+ * @param keyField identity field used to match the animation row (defaults to SERIES_ID)
+ */
+export const getHoverFractionSignal = (name: string, keyField: string = SERIES_ID): string => {
+    const fractionData = `data('${name}_hoverFractionData')`;
+    const lookup = `indexof(pluck(${fractionData}, '${keyField}'), datum.${keyField})`;
+    // default to the neutral emphasis level when this datum has no animation row
+    return `(${fractionData}[${lookup}] || {fraction: ${HOVER_NEUTRAL_TARGET}}).fraction`;
+};
+
+// The two ramps below take an emphasis-level expression (from getHoverFractionSignal — 0 = deemphasized,
```
c-lamoureux: This is good detail to have, but we should attach it to the function signatures. Having it separated here increases the risk that it becomes detached from the relevant code if the functions end up being moved/refactored.

c-lamoureux:
Do mid animation changes cause us to go from the current value to the target at the same speed as the start value to the target? Let's say we hover a line and it takes 3 seconds to go from 1 to 0.2 opacity. If I unhover the line at 1.5 seconds, is it going to take 1.5 seconds to go from 0.45 opacity back to 1, or is it going to take 3 seconds?

planning/research/hover-animation-system/hover-animation-system.md
```diff
@@ -0,0 +1,364 @@
+# Hover Animation System — Design & Architecture
```
c-lamoureux, line 1: Glad we have this documented! We'll need to make sure to keep it updated as we make decisions during this PR review and future changes as we run into new requirements/issues.

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
logan-a-wood: Vega doesn't have a keyed access function usable inline inside an expression. Using data() just returns a regular array and the only tools for looking through it are indexof and pluck.

packages/react-spectrum-charts-s2/src/stories/Line/Features/HoverAnimation/LineHoverAnimation.story.tsx
logan-a-wood: The plan for once everything is wired in was to have opacity and stroke width utilizing the animation system. Although, we could try only animating opacity and hook up stroke width in a future PR if that would be better.

c-lamoureux: I would split them into 2 PRs so we can track the features more easily in the git history.

packages/vega-spec-builder-s2/src/marks/hoverAnimationUtils.ts
c-lamoureux: Yeah you're right. Thanks!

packages/constants/constants.ts
```diff
@@ -165,6 +167,19 @@ export const PADDING_RATIO = 0.4;
 export const LINEAR_PADDING = 0;
 export const TRELLIS_PADDING = 0.2;
 
+// hover animation constants
+/** Timer signal update interval in ms. Caps timer signal update at ~60fps. */
```
c-lamoureux:
```suggestion
/** Timer signal update interval in ms. Caps timer signal update at ~60fps. 16 * 60 (frames) = ~1000ms*/
```

planning/research/hover-animation-system/legendHoverExample.vg.json
```diff
@@ -0,0 +1,1566 @@
...
+                    "target": 1
+                },
+                {
+                    "rscSeriesId": "Add Bar viz",
+                    "startTime": 0,
+                    "startValue": 1,
+                    "target": 1
+                }
+            ],
+            "on": [
```
c-lamoureux, line 732: This could be a potential performance hit since each of these are reading the same trigger so we're duplicating execution. I'm not sure what the impact would be and if there's a way we can reduce triggers by using a dynamic index as opposed to hard coding them. Something to keep in mind.

c-lamoureux:
The spec works really well. Experience feels smooth and it's relatively easy to manage and understand for what we're accomplishing with it. Really great work!

c-lamoureux:
Documenting some of the topics we discussed regarding performance:
- Creating extra data tables like this could create some scaling performance issues. We'll need to test that.
- If any of these data tables are constantly updating alongside the timer, even when values haven't changed, resource usage could quickly become problematic.

c-lamoureux:
It looks like the Storybook stories aren't able to load. The page crashes before the stories render. Especially seeing this on the hoverAnimation Docs section. May be related to some of the performance callouts in my previous comment.

logan-a-wood:
> Do mid animation changes cause us to go from the current value to the target at the same speed as the start value to the target? Let's say we hover a line and it takes 3 seconds to go from 1 to 0.2 opacity. If I unhover the line at 1.5 seconds, is it going to take 1.5 seconds to go from 0.45 opacity back to 1, or is it going to take 3 seconds?

Currently, no matter the start value (whether a line is mid animation when it's target value changes), it should take the same amount of time to animate. In your example, hovering off the line 1.5 seconds into the animation would cause it to take 3 seconds to go from 0.45 opacity back to 1.

This is a good catch since this seems like unwanted behavior. I'll fix the lerping logic to scale duration based on the distance it's animating.

logan-a-wood:
> It looks like the Storybook stories aren't able to load. The page crashes before the stories render. Especially seeing this on the hoverAnimation Docs section. May be related to some of the performance callouts in my previous comment.

I had previously made one of the stories really really big for performance testing. This is most likely the reason for Storybook not working. I have since then reduced the size of the performance testing stories.

c-lamoureux (approved):

PR #844, opened 2026-07-21
feat: AN-456589 Highlight bar on axis label hover

tanugarg04:
## Description

Implements [AN-456589](https://jira.corp.adobe.com/browse/AN-456589): hovering an axis label now highlights the corresponding bar(s) — the whole stack for stacked bars, the whole group for dodged bars — the same way hovering the bar itself already does. Other axis labels dim in sync with the bars. **S2 only**, per ticket scope.

- New Storybook stories: `AxisLabelHighlight` under `Bar`, `Stacked Bar`, and `Dodged Bar` features.

Screenshot 2026-07-21 at 3 52 01 PM

## Types of changes

## Checklist:

### DIFF
--- a/packages/vega-spec-builder-s2/src/chartInspect/chartInspectUtils.ts
+++ b/packages/vega-spec-builder-s2/src/chartInspect/chartInspectUtils.ts
@@ -231,15 +231,14 @@ export const addHoverdDimenstionAreaOpacityRules = (
   opacityRules: ({ test?: string } & NumericValueRef)[],
   markOptions: InspectParentOptions
 ) => {
-  if (!hasInspectWithDimensionAreaTarget(markOptions.chartInspects) || !('dimension' in markOptions)) return;
+  // only bar marks create the dimensionHoverArea signal/rect this rule depends on (see barSpecBuilder.ts,
+  // stackedBarUtils.ts, dodgedBarUtils.ts) - other mark types with a `dimension` field (e.g. scatter)
+  // must not get this rule since the signal it references would never exist for them.
+  if (markOptions.markType !== 'bar' || !('dimension' in markOptions)) return;
   const { name, dimension } = markOptions;
   const hoveredItemSignal = `${name}_${DIMENSION_HOVER_AREA}_${HOVERED_ITEM}`;
   opacityRules.push({
     test: `isValid(${hoveredItemSignal})`,
     signal: `${hoveredItemSignal}.${dimension} === datum.${dimension} ? 1 : ${FADE_FACTOR}`,
   });
-};
-
-export const hasInspectWithDimensionAreaTarget = (chartInspects: ChartInspectOptions[]) => {
-  return chartInspects.some(({ targets }) => targets?.includes('dimensionArea'));
 };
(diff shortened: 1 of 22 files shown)


packages/vega-spec-builder-s2/src/chartInspect/chartInspectUtils.ts
```diff
@@ -231,15 +231,14 @@ export const addHoverdDimenstionAreaOpacityRules = (
   opacityRules: ({ test?: string } & NumericValueRef)[],
   markOptions: InspectParentOptions
 ) => {
-  if (!hasInspectWithDimensionAreaTarget(markOptions.chartInspects) || !('dimension' in markOptions)) return;
+  // only bar marks create the dimensionHoverArea signal/rect this rule depends on (see barSpecBuilder.ts,
+  // stackedBarUtils.ts, dodgedBarUtils.ts) - other mark types with a `dimension` field (e.g. scatter)
+  // must not get this rule since the signal it references would never exist for them.
+  if (markOptions.markType !== 'bar' || !('dimension' in markOptions)) return;
```
c-lamoureux, line 235: Do we need to make this change? It might be true that only bar marks are using it today, but the existing version is more extensible.

packages/vega-spec-builder-s2/src/types/specUtil.types.ts
```diff
@@ -43,6 +43,7 @@ export type UserMeta = {
   chartOrientation?: Orientation;
   metricAxisCount?: number;
   patches?: Partial<Config>[];
+  barDimensionFields?: { name: string; dimension: string }[];
```
c-lamoureux: We want to avoid UserMeta usage whenever possible so it doesn't become bloated. Since AxisSpecBuilder runs after BarSpecBuilder, is there any way we can get the information we need via the spec itself in the AxisSpecBuilder?

tanugarg04: In `packages/vega-spec-builder-s2/src/types/specUtil.types.ts:22-36` has a comment justifying usermeta for this exact scenario:

"A chart element like bar needs to communicate information about its orientation to the axes elements. However, axis elements are built within their own axisSpecBuilder, which is independent from the barSpecBuilder. Parsing the spec for information about the bar is brittle, since implementation details may change. A preferred approach would be to add the needed information to usermeta."

That's the reason I followed this pattern (mirrors usermeta.chartOrientation/usermeta.interactiveMarks, which solve the identical cross-builder problem). 

We should update the comment if userMeta is not the way to go.
I would look into another approach.

packages/vega-spec-builder-s2/src/chartInspect/chartInspectUtils.ts
tanugarg04: Good question — I checked, and the guard is necessary, not optional. Here's concrete proof:

Scatter already has a dimension field (scatterSpec.types.ts:51) and already reaches this exact function through its own path: getOpacity() in scatterMarkUtils.ts:131-137 calls addHoveredItemOpacityRules(rules, scatterOptions) directly (a separate call site from getMarkOpacity, which is typed to Bar/Donut/Venn only — Scatter has its own opacity function that isn't type-restricted the same way). It's gated by the same isInteractive(scatterOptions) check bar uses.

Scatter never creates a ${name}_dimensionHoverArea_hoveredItem signal anywhere in scatter/. So with the markType !== 'bar' guard removed, any interactive Scatter chart (e.g. one with a ChartPopover) that also sets dimension would get { test: isValid(scatter0_dimensionHoverArea_hoveredItem), ... } spliced into its opacity rules — referencing a signal that doesn't exist anywhere in the spec.

Error depicting in story-
Screenshot 2026-07-22 at 1 41 40 PM

That's not a silent no-op — it's a hard parse/runtime failure, which would crash the entire chart, not just disable one opacity rule.

c-lamoureux (approved):

PR #861, opened 2026-08-04
feat(s2): AxisThumbnail follows the diverging axis to the zero line

tanugarg04:
feat(s2): AxisThumbnail follows the diverging axis to the zero line
Jira: [AN-462759](https://jira.corp.adobe.com/browse/AN-462759)

## Description

With `<Bar diverging>`, the dimension axis and its labels move to the opposing scale's zero line, but axis thumbnails stayed pinned at the chart edge. Thumbnails are rendered as top-level sibling image marks (not nested in the axis group), so they never picked up the axis's diverging offset — leaving them visually detached from the labels they belong to.

Thumbnail image marks (and the label's thumbnail-clearance offset) now flip per row to the opposite side of the zero line from each row's own bar, mirroring the diverging label align flip

## How Has This Been Tested?

react-spectrum-charts-s2/.../Bar/Diverging.story.tsx — HorizontalWithThumbnail + VerticalWithThumbnail stories

Screenshot 2026-08-04 at 4 29 03 PM

## Types of changes

## Checklist:

### DIFF
--- a/packages/vega-spec-builder-s2/src/chartSpecBuilder.ts
+++ b/packages/vega-spec-builder-s2/src/chartSpecBuilder.ts
@@ -230,8 +230,8 @@ export function buildSpec({
     spec.axes = [...spec.axes].sort((a, b) => Number(isDivergingAxis(a)) - Number(isDivergingAxis(b)));
   }
 
-  // diverging thumbnails are bandwidth-sized (width) and placed off the zero line (height), so Vega's default `pad`
-  // autosize feeds them back into width/height and collapses the plot; RSC sizes the view itself, so pin autosize off
+  // diverging thumbnails anchor to the opposing scale's zero line, so their overflow is coupled to the fitted plot
+  // size and `fit` has no stable fixpoint — it re-fits per pulse and collapses on hover; RSC sizes the view, so opt out
   if (spec.usermeta?.divergingBarMarks?.length && axes.some((axis) => axis.axisThumbnails?.length)) {
     spec.autosize = { type: 'none' };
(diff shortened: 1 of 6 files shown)


packages/vega-spec-builder-s2/src/chartSpecBuilder.ts
```diff
@@ -230,6 +230,12 @@ export function buildSpec({
     spec.axes = [...spec.axes].sort((a, b) => Number(isDivergingAxis(a)) - Number(isDivergingAxis(b)));
   }
 
+  // diverging thumbnails are bandwidth-sized (width) and placed off the zero line (height), so Vega's default `pad`
+  // autosize feeds them back into width/height and collapses the plot; RSC sizes the view itself, so pin autosize off
+  if (spec.usermeta?.divergingBarMarks?.length && axes.some((axis) => axis.axisThumbnails?.length)) {
+    spec.autosize = { type: 'none' };
+  }
+
```
tanugarg04: Lets confirm this fix before merging.

c-lamoureux: This will cause a break in other sizing calculations in the chart. It looks like the shrinking bug is in main, I'm only seeing it locally though, not in the hosted storybook. I believe the way we're calculating the offset and the chart size is resulting in a circular size change until the width hits 0. Not sure why interaction is triggering it yet. 

Let's take this out and come up with a more comprehensive fix.

c-lamoureux (approved):

PR #822, opened 2026-06-09
feat: initial data-navigator architecture

c-lamoureux:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](https://opensource.adobe.com/cla.html).
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #870, opened 2026-08-07
feat: Line keyboard nav Enhance S2 line features, reference lines, and documentation updates

majornista:
## Description

Adds accessible keyboard navigation ("data-navigator") support for S2 line charts, bringing lines to parity with the existing bar navigation support, and wires that navigation into `ChartPopover`/`ChartInspect` for both Line and Bar. Specifically:

**Core line navigation structure**
- **`buildLineStructure.ts`** — builds the navigator's line/point node structure (series → points), including focus node IDs so keyboard navigation can address individual points and whole lines.
- **`segmentId.ts`** — shared helper for constructing unique navigation node IDs.
- **`lineFocusRingUtils.ts`** / **`chartFocusRingUtils.ts`** — generate the chart-level, line-level, and point-level focus ring marks rendered in the Vega spec when a node is focused via keyboard. The chart-level ring was later extracted into a shared `chartFocusRingUtils.ts` since it was identical between Bar and Line.
- **`focusMatchUtils.ts`** — shared expression helper (`getFocusedGroupOrItemMatchExpr`) used to test whether a given datum matches the currently focused dimension/item, reused across the line marks and legend opacity logic. The match convention (prefix vs. suffix) is now an explicit parameter supplied by each caller rather than inferred, fixing a false-positive where a series/dimension name that happened to be a numeric prefix or suffix of another node's composite ID (e.g. a line literally named `"3"` colliding with another line's third point) could match the wrong node.
- **Click-to-focus** — clicking a line/point in the Vega chart now updates navigator focus state, so mouse and keyboard interactions stay in sync (`Navigator.tsx`, `dataNavigatorAdapter.ts`). Bars are now interactive when accessible navigation is enabled, so they participate in the same click-to-focus wiring as lines.
- New constants (`NAVIGATION_INDEX_FIELD` and related) for keying a point's position within its line for keyboard-navigation focus, computed in JS and passed into the spec via a signal rather than a per-row Vega `formula`/`window` transform (avoids an O(N²) cost and avoids silently reordering the table).

**Bar keyboard-navigation parity & orientation support**
- StackedBar gains its own keyboard navigation rules (drill into a stack's segments, sibling/parent/child edges), reaching parity with the line structure above.
- Keyboard navigation is now orientation-aware: horizontal Bar/StackedBar charts swap `ArrowLeft`/`ArrowRight` ↔ `ArrowUp`/`ArrowDown` so the keys always match the chart's actual visual layout instead of assuming a vertical chart (`getBaseNavigationRules`/`getStackedBarNavigationRules` in `buildBarStructure.ts`). The generated ARIA description text swaps accordingly.
- A StackedBar's dataNavigator label now includes the summed metric total for that stack (e.g. "Chrome. Contains 3 bars, 10 Downloads."), resolved against whichever `Axis` sits on the metric side of the chart so the label uses that axis's real title instead of the raw data field name.

**Focus/hover deemphasis correctness**
- Lines, bars, and legend entries dim non-focused series/points when a node is focused via keyboard (mirroring the existing hover-deemphasis), but only when the chart already has some other hover-capable interactive feature (`onClick`, `ChartPopover`, `ChartInspect`, or a `highlightedItem` prop) — `accessibleNavigation` on its own no longer introduces opacity changes that mouse hover wouldn't already produce.
- An `interactionModality` signal tracks whether pointer or keyboard was most recently used, so hover-driven emphasis (opacity, stroke width, hover labels) and keyboard focus don't fight each other — including restoring the correct look when the pointer leaves the chart or moves onto a legend entry, and preventing a stale mouse-hover rule from holding an entire StackedBar column at full opacity after moving keyboard focus with the arrow keys (it now correctly dims down to just the focused segment).

**Keyboard-driven popover and tooltip integration**
- Enter/Space on a keyboard-focused point now opens its `ChartPopover` the same way a click does, via a shared `selectAndOpenPopover` helper extracted from the existing click handler (`markClickUtils.ts`) — and now also unconditionally fires the mark's own `onClick`, so click handlers behave identically whether the mark was activated by mouse or keyboard.
- Keyboard activation and focus-driven tooltips now mirror real click/hover gating exactly: a popover only opens, and a `ChartInspect` tooltip only appears, when that feature is actually declared on the chart for the focused mark. Previously, `accessibleNavigation` alone could pop open an undeclared `ChartPopover` on Enter/Space, or fall back to vega-tooltip's own default all-fields table on focus when no `ChartInspect` was configured.
- A keyboard-focused point drives the `ChartInspect` tooltip the same way mouse hover does, using scale-projected geometry (`focusedItemGeometry.ts`) since a keyboard-focused datum has no real Vega `Item` to read bounds from.
- Escape has layered, precedence-ordered behavior: dismiss a visible `ChartInspect` tooltip first, then close an open `ChartPopover` (with a short grace window so a fast double-Escape right after closing isn't read as a separate drill-out request), then drill out a navigation level.
- Closing the popover restores keyboard focus to the point that had it (double-`requestAnimationFrame`, timed to run after React Aria's own single-rAF focus restore).

**Bug fixes found and fixed along the way**
- Bar charts with `accessibleNavigation` enabled and no other interactive feature (no popover/inspect/onClick) never got an `interactionModality` signal registered, so the first keyboard focus event threw a Vega runtime error and silently broke navigation entirely.
- Multi-mark charts (2+ Bar and/or Line marks) with `accessibleNavigation` enabled pushed duplicate `focusedItem`/`focusedRegion`/`focusedDimension` signals and duplicate `chartFocusRing` marks into the spec, which Vega rejects as a duplicate declaration. Fixed by centralizing these chart-wide primitives (`addFocusSignals`, `addChartFocusRing`) with existence guards, reused across the chart/bar/line spec builders, and by registering the focus signals once at the chart level (`getDefaultSignals`) rather than per-mark — which also fixes a related crash on any chart with `accessibleNavigation` and a legend but no Bar/Line mark.
- Keyboard activation and focus-driven tooltip resolution keyed off the mark's raw JSX `name` prop, which is `undefined` for every Bar (no default) and any Line without an explicit `name` (its `name = 'line0'` default never runs, since Line is a render-null placeholder component). Fixed by resolving the mark's name in `RscChart` the same way the spec builder itself does.
- The global `Escape`-dismisses-`ChartInspect`-tooltip listener was firing for every chart using `ChartInspect`, not just ones with `accessibleNavigation` enabled; it's now gated behind that flag.

**Cleanup and docs**
- Extracted a shared `nodeSemanticsUtils.ts` (`applyDefaultLabels`) to de-duplicate identical accessible-label logic between `buildBarStructure.ts` and `buildLineStructure.ts`.
- Extracted `graphEdgeUtils.ts` (`addSiblingEdge`/`addOneSidedEdge`) and a shared `dataNavigatorTestUtils.ts` `move()` test helper, de-duplicating logic that had been copy-pasted between `buildBarStructure.ts` and `buildLineStructure.ts` (and their tests).
- Extracted a shared `hasAccessibleNavigation` helper (`markUtils.ts`), replacing three separate inline derivations of the same `accessibleNavigation` check.
- Added a `useLiveRef` pattern in `Navigator.tsx` so parent re-renders that change callback props don't tear down and rebuild the whole navigation structure.
- New "Accessible keyboard navigation" section in `packages/docs/docs/spectrum2/overview.md` (key bindings table, popover/inspect behavior, orientation-swap note), with pointers from `line.md` and `bar.md`.
- `LineAccessibleNavigation.story.tsx` moved into a `Features/AccessibleNavigation/` subdirectory to match this repo's multi-story Storybook convention, and now also exercises `ChartPopover` alongside `ChartInspect`.
- Storybook coverage added via `LineAccessibleNavigation.story.tsx`, plus unit tests for all new/modified logic (`buildLineStructure`, `lineFocusRingUtils`, `chartFocusRingUtils`, `focusMatchUtils`, `lineDataUtils`, `lineMarkUtils`, `lineSpecBuilder`, `legendUtils`, `chartInspectUtils`, `dataNavigatorAdapter`, `Navigator`, `focusedItemGeometry`, `nodeSemanticsUtils`, `signalSpecBuilder`, `chartSpecBuilder`, `useNewChartView`, `buildBarStructure`, `buildChartStructure`, `barSpecBuilder`), including targeted regression tests for the orientation swap, metric-total labels, and the interactivity-gating fixes above.

## Motivation and Context

S2 line charts had no keyboard-accessible way to navigate between series and data points — only bars supported the data-navigator. This closes that gap so line charts meet the same accessibility bar, brings Bar/StackedBar navigation to orientation-aware parity, wires keyboard navigation into the existing popover/tooltip interactions so keyboard users get the same information and click behavior mouse users do, and fixes several correctness bugs (some pre-existing, some introduced along the way) that could silently break keyboard navigation, crash the chart render, or introduce visual side effects (opacity dimming, unrequested popovers/tooltips) that `accessibleNavigation` shouldn't cause on its own.

## How Has This Been Tested?

- Added unit tests covering the new line structure builder, focus ring generation (chart/line/point level), the focus-match expression helper, the updated legend/line opacity and mark-building logic, the popover/tooltip activation path, the signal/mark-deduplication fixes for multi-mark charts, orientation-aware bar navigation, and the metric-total label computation.
- Added targeted regression tests for each bug fix, including a real end-to-end test (mounting a chart with no explicit mark `name` and a `ChartPopover`) confirming Enter-to-activate opens it, a test confirming a second navigable mark doesn't produce duplicate signals/marks, and tests confirming keyboard focus/activation produce no opacity dimming, no default tooltip, and no popover when the chart doesn't declare those features.
- Ran `yarn test` across the touched packages (`vega-spec-builder-s2`, `react-spectrum-charts-s2`) — all suites passing — plus `yarn lint` and `yarn tsc --noEmit` clean.
- Verified interactively in Storybook via the `LineAccessibleNavigation` stories (now including `ChartPopover`) and the Bar `AccessibleNavigation`/`StackedBar` stories: keyboard navigation between lines/points/stacks in both orientations, focus rings rendering at every level, legend/line deemphasis gated to real interactivity, Enter/Space firing `onClick` and opening popovers only when configured, and the Escape dismiss/close/drill-out precedence.
- This branch was run through two rounds of multi-persona code review (security/correctness/standards/architecture/devops/documentation), each with a false-positive-elimination pass; all accepted findings were addressed.

## Types of changes

## Checklist:


c-lamoureux:
This is great! I'll perform a full review by the end of the week so we can get it in. Thanks for taking the time to put it together!

majornista:
> This is great! I'll perform a full review by the end of the week so we can get it in. Thanks for taking the time to put it together!

Note that the huge number of commits and files changed comes from merging `main` into `dataNavigationManual` before adding the relevant commit: f1fb6801647d5b2267cbce27e71f46fca126638e.

majornista:
I need to work on the behavior with chart tooltips, so that they display and position relative to the focused node, and chart popovers, so that they can be toggled from the data navigator node.

c-lamoureux:
I'm still reviewing. I'd like to see what the feedback from Frank is on #822 as well before we bring this in.

PR #889, opened 2026-08-19
feat: s2 donut summary, metric value/label sizing to named size tiers

madelineluke:
## Description

Replaces DonutSummary's placeholder ratio-based font sizing (`donutRadius * holeRatio * 0.35`, with the label always forced to exactly half the value's size) with two independent threshold scales/signals (value, label) keyed off the donut's outer diameter. A diameter snaps to the nearest of the five named S2 size tiers (XS/S/M/L/XL) via shared midpoint cutpoints, rather than the old scale's snap-down behavior. Value/label font sizes are now independently specified per tier (matching the design tokens), not derived as a fixed ratio of each other. Text truncation/fit-based visibility (`DONUT_SUMMARY_MIN_RADIUS`) intentionally continues to use the real `holeRatio`-derived inner radius, not the tier's fixed ring-width token - the actual arc ring geometry (`getArcMark`) is untouched by this change and still `holeRatio`-based, so truncation math stays matched to what's actually rendered. That ring-geometry/design-token mismatch is real and confirmed (measured against the Figma reference) but is tracked as a separate issue rather than folded into this PR (see #891). Adds a new non-pre-alpha `Donut/Features/Donut Summary` Storybook story showing all five size tiers side by side.

## Related Issue

Implements `planning/specs/donut/donut-summary-responsive-sizing.json` - second spec in the `planning/specs/donut/` s2 donut feature series (reordered ahead of `donut-direct-labels`, which depends on the size-tier constants this PR establishes). No GitHub issue tracked.

## Motivation and Context

The center metric value/label sizing didn't match the S2 design system's named size tiers, and hardcoded the label to always be exactly half the value's font size, which doesn't hold once independently-specified per-tier sizes are used.

## How Has This Been Tested?

- `yarn test --testPathPattern="react-spectrum-charts-s2|vega-spec-builder-s2"` - 2112/2112 passing
- `yarn lint` - clean
- `yarn tsc --noEmit` - no new errors (pre-existing unrelated errors confirmed present on `main`)
- Verified visually in Storybook (`Donut/Features/Donut Summary/Size Tiers`) that value/label font sizes scale correctly across all five tiers

N/A - verified in Storybook, see test plan above.

## Types of changes

## Checklist:


packages/vega-spec-builder-s2/src/donut/donutSummaryUtils.ts
```diff
@@ -119,11 +127,16 @@ export const getDonutSummarySignals = (donutOptions: DonutSpecOptions): Signal[]
...
   }
-  const { name: donutName, holeRatio } = donutOptions;
+  const { name } = donutOptions;
+  const donutDiameter = `2 * ${DONUT_RADIUS}`;
   return [
     {
-      name: `${donutName}_summaryFontSize`,
-      update: `scale('${donutName}_summaryFontSizeScale', ${DONUT_RADIUS} * ${holeRatio} * ${DONUT_SUMMARY_FONT_SIZE_RATIO})`,
+      name: `${name}_summaryValueFontSize`,
+      update: `scale('${name}_summaryValueFontSizeScale', ${donutDiameter})`,
```
c-lamoureux, line 135: Why did we base this on the diameter rather than the chart width?

c-lamoureux (dismissed):

PR #894, opened 2026-08-26
feat(donut-s2): Direct Labels - base implementation, positioning and scaling 

madelineluke:
## Description

Implements the ring-gap sizing/hemisphere-anchor portion of the `donut-direct-labels` spec, and fixes a real resize-reactivity bug found while building the responsive Storybook example for it.

- Adds `getDonutOuterRadiusExpr` to deterministically reserve room for the direct-label ring gap and hemisphere pull-back within the donut's own radius, so the ring and its labels always fit within a given container size in a single pass, without relying on Vega's `autosize` growing beyond it. This is threaded through the arc mark, the empty-state ring, and `DonutSummary`'s radius/font-size/truncation calculations so they all agree on the same effective radius.
- Fixes a bug where `radius`/`dx`/`fontSize`/`dy`/`align`/`baseline` on the segment label text marks lived in Vega's `enter` encode set, which only evaluates once when a mark is first created. On resize, only the shared `x`/`y` anchor (in `update`) tracked the new container size, so labels were uniformly translated by the container's center shift instead of properly rescaling - causing them to visually detach from (or overlap into) the ring at certain sizes. Moved these encodings into `update`.
- Replaces the CSS-resize-based Storybook example with a width-slider-driven `Responsive` story (matching the existing `Line` `DirectLabelSizeScaling` pattern, with visible size-tier breakpoint markers), plus a `ManySegmentsResponsive` variant using the existing `sliveredDonutData` fixture to stress-test label crowding with 15 segments.

## Related Issue

`planning/specs/donut/donut-direct-labels.json` (approved)

## Motivation and Context

Direct labels on a donut need a ring gap and can pull back horizontally to avoid overlapping the ring near the 9/3 o'clock cardinal points. Without reserving space for that up front, labels could end up demanding more room than the container had, either forcing the chart to grow beyond its given size or causing the ring to shrink disproportionately. Fixing the resize-reactivity bug was necessary because the size-tier-driven font/anchor calculations only take effect on the very first render otherwise, making the ring and its labels visibly drift apart on any subsequent resize.

## How Has This Been Tested?

- Unit tests updated in `segmentLabelUtils.test.tsx` and `donutSummaryUtils.test.tsx` for the new radius-reservation formula and the `enter`→`update` encode move.
- Full `vega-spec-builder-s2`/`react-spectrum-charts-s2` test suite passes (131 suites, 2125 tests).
- `yarn lint` and `yarn tsc --noEmit` are clean (no new errors; two pre-existing unrelated failures confirmed present on the base commit too).
- Manually verified in Storybook via Playwright screenshots across the full slider range (80px-500px, XS-XL tiers) on both the `Responsive` and `ManySegmentsResponsive` stories: the ring and its labels now scale together in real time with no overlap, including immediately after a resize (previously frozen from the initial render).

See the `Responsive` and `ManySegmentsResponsive` stories under `React Spectrum Charts 2/Donut/Features/Direct Label` in Storybook.

## Types of changes

## Checklist:

### DIFF
--- a/llm/skills/s2-donut/donut-chart-tokens/SKILL.md
+++ b/llm/skills/s2-donut/donut-chart-tokens/SKILL.md
@@ -105,14 +105,16 @@ Sizes marked "derived" were computed rather than individually measured. S/L (rel
 | Metric total | `18px` (derived) | `20px` | `22px` | `36px` | `50px` (derived) | `800` | `text-primary` |
 | Metric label | `12px` (derived) | `14px` | `16px` | `20px` | `24px` (derived) | `700` | `text-primary` |
 | Metric delta line | `12px` (derived) | `14px` | `16px` | `20px` | `24px` (derived) | `800` | `sentiment-positive` / `sentiment-negative` |
-| Direct label — segment name | `9px` (derived) | `10.5px` (derived) | `12px` | `15px` (derived) | `18px` (derived) | `400` | `gray-700` |
-| Direct label — value | `12px` (derived) | `14px` (derived) | `16px` | `20px` (derived) | `24px` (derived) | `700` | `gray-700` at rest → segment's own categorical color on hover (see §5) |
+| Direct label — segment name | `7.5px` (derived) | `9px` (derived) | `10.5px` (derived) | `12px` | `15px` (derived) | `400` | `gray-700` |
+| Direct label — value | `10px` (derived) | `12px` (derived) | `14px` (derived) | `16px` | `20px` (derived) | `700` | `gray-700` at rest → segment's own categorical color on hover (see §5) |
 | Advanced label — segment name | same size as Direct label — value | | | | | `400` | `gray-700` |
 | Advanced label — value/% | `13.5px` (derived) | `15.75px` (derived) | `18px` | `22.5px` (derived) | `27px` (derived) | `800` | `gray-800` at rest → segment's own categorical color on hover (see §5) |
 | Advanced label — detail row | `8.25px` (derived) | `9.6px` (derived) | `11px` | `13.75px` (derived) | `16.5px` (derived) | `400` | `gray-700` |
 | Benchmark label — single-line ("Target" only) | `12px` (derived) | `14px` | `16px` | `20px` | `24px` (derived) | `700` | `gray-800` |
 | Benchmark label — two-line, label part | same as Direct label — segment name | | | | | `400` | `gray-700` |
 | Benchmark label — two-line, value part | same size as Direct label — value | | | | | `700` | `gray-800` at rest → segment's own categorical color on hover (see §5) |
+
+**Direct label correction:** the Direct label rows above were originally computed with the generic `×0.875/×1/×1.25`-from-M method described above, which was wrong — confirmed against the actual design tokens, each named tier's Direct label size is one step smaller than that method produced (new tier N = old tier N−1's value; e.g. current M's `12px`/`16px` is the old S value). `XS` is still derived by extending the corrected S→M step one increment further down (`XS = S − (M−S)`).
 
 ---
 
--- a/packages/react-spectrum-charts-s2/src/stories/Donut/Features/DirectLabel/DirectLabels.story.tsx
+++ b/packages/react-spectrum-charts-s2/src/stories/Donut/Features/DirectLabel/DirectLabels.story.tsx
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+import { ReactElement, useState } from 'react';
+
+import { StoryFn } from '@storybook/react';
+
+import { DONUT_SIZE_TIER_CUTPOINTS } from '@spectrum-charts/constants';
+import { ChartData } from '@spectrum-charts/vega-spec-builder-s2';
+
+import { Chart } from '../../../../Chart';
+import useChartProps from '../../../../hooks/useChartProps';
+import { Donut, SegmentLabel } from '../../../../pre-alpha';
+import { bindWithProps } from '../../../../test-utils';
+import { SegmentLabelProps } from '../../../../types';
+import { basicDonutData, sliveredDonutData } from '../../../components/Donut/data';
+
+export default {
+  title: 'React Spectrum Charts 2/Donut/Features/Direct Label',
+  component: SegmentLabel,
+};
+
+const CHART_SIZE = 400;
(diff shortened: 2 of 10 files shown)


packages/react-spectrum-charts-s2/src/stories/Donut/Features/DirectLabel/DirectLabels.story.tsx
```diff
@@ -0,0 +1,204 @@
...
+  component: SegmentLabel,
+};
+
+const THUMB_HEIGHT = 32;
+
+// Mirrors getDonutOuterRadiusExpr's reservation math (donutUtils.ts): since a SegmentLabel is
+// always present here, the donut's actual rendered radius - and therefore its font-size/ring-width
+// size tier - is based on a smaller, label-reserved diameter, not the raw container width.
+const getEffectiveDiameter = (containerWidth: number): number => {
+  const rawRadius = containerWidth / 2 - 2;
```
c-lamoureux, line 41: Why - 2? Just want to make sure it's clear what we're doing here.

packages/react-spectrum-charts-s2/src/stories/Donut/Features/DirectLabel/DirectLabels.story.tsx
```diff
@@ -0,0 +1,204 @@
...
+  const reservedRadius = (rawRadius - DONUT_LABEL_RING_GAP) / (1 + DONUT_LABEL_MAX_ANCHOR_OFFSET_RATIO);
```
c-lamoureux, line 42: Same for 1 +, what is it for?

packages/react-spectrum-charts-s2/src/stories/Donut/Features/DirectLabel/DirectLabels.story.tsx
```diff
@@ -0,0 +1,204 @@
...
+  return 2 * reservedRadius;
+};
+
+// Inverse of getEffectiveDiameter - the container width at which the effective (label-reserved)
+// diameter reaches a given tier cutpoint, so breakpoint markers land at the right slider position.
+const getContainerWidthForDiameter = (diameter: number): number =>
+  diameter * (1 + DONUT_LABEL_MAX_ANCHOR_OFFSET_RATIO) + 4 + 2 * DONUT_LABEL_RING_GAP;
```
c-lamoureux, line 49: I think we should be a bit more clear about where the fixed values in this formula are coming from.

packages/vega-spec-builder-s2/src/donut/segmentLabelUtils.ts
```diff
@@ -48,6 +73,118 @@ const applySegmentLabelPropDefaults = (
...
+  for (let i = rules.length - 2; i >= 0; i--) {
+    const rule = rules[i] as { test?: string } & TextValueRef;
+    expr = rule.test ? `${rule.test} ? (${getValue(rule)}) : (${expr})` : getValue(rule);
+  }
+  return expr;
+};
+
+/**
+ * Gets the shared anchor radius for a label's name/value lines - always the ring-gap point. The
+ * hemisphere-mirrored pull-back lives entirely in getLabelAnchorDxExpr as a horizontal pixel offset,
```
c-lamoureux: I'm finding this comment difficult to follow. What does hemisphere-mirrored pull-back mean here?

packages/vega-spec-builder-s2/src/donut/segmentLabelUtils.ts
```diff
@@ -48,6 +73,118 @@ const applySegmentLabelPropDefaults = (
...
+ * @param donutOptions
+ * @returns vega expression string
+ */
+const getLabelAnchorRadiusExpr = (donutOptions: DonutSpecOptions): string =>
+  `${getDonutOuterRadiusExpr(donutOptions)} + ${DONUT_LABEL_RING_GAP}`;
+
+/**
+ * Gets the shared horizontal pixel offset for a label's name/value lines. Right-hemisphere labels get
+ * no offset (anchor sits at the ring-gap point and grows away from the ring). Left-hemisphere labels
+ * get pulled left by the wider line's real rendered width, so that line's far (right) edge lands
```
c-lamoureux: I'm having trouble following this as well.

c-lamoureux: Oh, is it referring to the text line properties?

c-lamoureux:
We'll probably address this in a later PR but commenting here since it's relevant. We'll have to do a design review for how we're handling gap scaling. I don't have a preferred solution at this time. Maybe we could group small segments and not separate them with a gap at all. 

We're reducing the arc by 1/3 for gap when it's below a threshold size, rather than by a fixed amount. A potential problem with this is that, proportionally, larger segments visually take up a larger share of the Donut than smaller ones (40 + 5 = 45 segment + gap, 4 + 2 = 6 segment + gap. The 6 segment is reduced by 33% but the 45 segment is reduced by 11.1%). Potential data representation inaccuracy.

packages/vega-spec-builder-s2/src/donut/segmentLabelUtils.ts
madelineluke: This comment is incredibly confusing and needs to be rewritten. This expression is simply getting the value where the label is initially anchored, before any collision calculations happen. The "hemisphere-mirrored pull-back:=" was trying to say that instead of the the text growing outwards towards the right, on the left hand side the text is pulled towards the left. Horrible lol 

I will fix this comment

packages/vega-spec-builder-s2/src/donut/segmentLabelUtils.ts
madelineluke: Yeah, also horribly written and confusing. This expression function is determining the horizontal offset for each label. It looks at which hemisphere it's on (right or left) and then sets the offset. The right side has a dx of 0, and the left side has a negative value. 

I will update this as well

packages/react-spectrum-charts-s2/src/stories/Donut/Features/DirectLabel/DirectLabels.story.tsx
madelineluke: I think claude was just trying to duplicate expressions that are already existing so that we can show the outer diameter breakpoints in the responsive story. I don't even know if all of this code is worth it, it's not technically necessary

llm/skills/s2-donut/donut-chart-tokens/SKILL.md
```diff
@@ -105,15 +105,17 @@ Sizes marked "derived" were computed rather than individually measured. S/L (rel
...
 | Advanced label — detail row | `8.25px` (derived) | `9.6px` (derived) | `11px` | `13.75px` (derived) | `16.5px` (derived) | `400` | `gray-700` |
 | Benchmark label — single-line ("Target" only) | `12px` (derived) | `14px` | `16px` | `20px` | `24px` (derived) | `700` | `gray-800` |
 | Benchmark label — two-line, label part | same as Direct label — segment name | | | | | `400` | `gray-700` |
 | Benchmark label — two-line, value part | same size as Direct label — value | | | | | `700` | `gray-800` at rest → segment's own categorical color on hover (see §5) |
 
+**Direct label correction:** the Direct label rows above were originally computed with the generic `×0.875/×1/×1.25`-from-M method described above, which was wrong — confirmed against the actual design tokens, each named tier's Direct label size is one step smaller than that method produced (new tier N = old tier N−1's value; e.g. current M's `12px`/`16px` is the old S value). `XS` is still derived by extending the corrected S→M step one increment further down (`XS = S − (M−S)`).
```
c-lamoureux, line 117: This seems like the wrong place for this info.

c-lamoureux (approved):

PR #895, opened 2026-08-27
feat(donut-s2): Hover - segment hover fade, direct-label color switch, and legend bidirectional highlight

madelineluke:
## Description

Implements the `donut-hover` spec: hovering a donut segment fades its siblings, switches the hovered segment's direct-label value to its own categorical color, and syncs highlight state bidirectionally with a paired `<Legend>`.

- Extends `isInteractive` so a donut with a `SegmentLabel` showing `value`/`percent` is hover-interactive on its own, with no popover/inspect required. Verified no other mark type declares a `segmentLabels` field, so this is a no-op everywhere else.
- Adds a `SERIES_ID` data transform (reusing the existing generic `getSeriesIdTransform` helper) whenever a donut is interactive or paired with a highlighting `Legend` - needed for both hover directions with a Legend to actually match rows, since donut's data never had this field before, unlike Line/Bar.
- Both `SegmentLabel` text marks now fade via the same `getMarkOpacity` mechanism the arc mark already used, and the value line switches to the hovered segment's own categorical color (the name line stays gray-700, unchanged) - matching the Figma reference exactly.
- Adds `legendHighlightSignals` to `DonutSpecOptions` and consumes it on the arc mark, fading non-matching segments when a paired Legend entry is hovered - the reverse direction of the arc's own hover already highlighting the legend generically via `userMeta.interactiveMarks`.
- Adds `Donut/Features/Hover` stories: a bare `SegmentLabel` case (no popover) and a `Legend`-paired case demonstrating both hover directions.

## Related Issue

`planning/specs/donut/donut-hover.json` (approved)

## Motivation and Context

Today a donut with only a direct label configured (no popover/inspect) produces no hover feedback at all, and even where hover exists (via a popover), the direct label's value never reflects which segment is hovered and never fades with its de-emphasized arc. Separately, a donut paired with a `<Legend>` never highlighted in either direction, unlike Line/Bar, because the generic `legendHighlightSignals` mechanism was already being spread into donut's options but silently dropped (untyped, unconsumed), and donut's data was missing the `SERIES_ID` field the legend's generic hover matching relies on.

## How Has This Been Tested?

- Unit tests added/updated in `markUtils.test.ts`, `donutSpecBuilder.test.ts`, `donutUtils.test.ts`, and a new `hover behavior` block in `segmentLabelUtils.test.tsx`.
- Full `vega-spec-builder-s2`/`react-spectrum-charts-s2` test suite passes (131 suites, 2120 tests).
- `yarn lint` and `yarn tsc --noEmit` are clean (no new errors; two pre-existing unrelated failures confirmed present on `main` too).
- Verified live in Storybook via Playwright, using precise on-pixel hover simulation (not bounding-box centers, which land inside a wedge's hole for thin rings): hovering an arc fades siblings to 0.2 opacity, keeps the hovered value's color, and highlights the matching legend entry; hovering a legend entry fades non-matching arcs. Both directions confirmed working end-to-end.

See the `SimpleDirectLabel` and `WithLegend` stories under `React Spectrum Charts 2/Donut/Features/Hover` in Storybook.

## Types of changes

## Checklist:

### DIFF
--- a/packages/vega-spec-builder-s2/src/donut/donutSpecBuilder.ts
+++ b/packages/vega-spec-builder-s2/src/donut/donutSpecBuilder.ts
@@ -22,6 +22,7 @@ import {
 } from '@spectrum-charts/constants';
 import { toCamelCase } from '@spectrum-charts/utils';
 
+import { getSeriesIdTransform } from '../data/dataUtils';
 import { isInteractive } from '../marks/markUtils';
 import { addFieldToFacetScaleDomain } from '../scale/scaleSpecBuilder';
 import { addHoveredItemSignal } from '../signal/signalSpecBuilder';
@@ -101,12 +102,19 @@ export const addDonut = produce<
 );
 
 export const addData = produce<Data[], [DonutSpecOptions]>((data, options) => {
-  const { name, isBoolean } = options;
+  const { color, legendHighlightSignals, name, isBoolean } = options;
   const filteredTableIndex = data.findIndex((d) => d.name === FILTERED_TABLE);
 
   //set up transform
   data[filteredTableIndex].transform = data[filteredTableIndex].transform ?? [];
   data[filteredTableIndex].transform?.push(...getPieTransforms(options));
+  // Needed for both hover directions with a paired Legend: this mark's own hover highlighting the
+  // legend (generic mark-hover loop in legendUtils.ts) and a legend hover fading this mark's arcs
+  // (getLegendHighlightOpacityRules in donutUtils.ts) - both match on datum[SERIES_ID], which donut's
+  // rows don't otherwise have, unlike Line/Bar.
+  if (isInteractive(options) || legendHighlightSignals?.length) {
+    data[filteredTableIndex].transform?.push(...getSeriesIdTransform([color]));
+  }
 
   if (isBoolean) {
     //select first data point for our boolean value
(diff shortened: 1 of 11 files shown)


packages/vega-spec-builder-s2/src/donut/donutSpecBuilder.ts
```diff
@@ -101,12 +102,19 @@ export const addDonut = produce<
...
+  const { color, legendHighlightSignals, name, isBoolean } = options;
   const filteredTableIndex = data.findIndex((d) => d.name === FILTERED_TABLE);
 
   //set up transform
   data[filteredTableIndex].transform = data[filteredTableIndex].transform ?? [];
   data[filteredTableIndex].transform?.push(...getPieTransforms(options));
+  // Needed for both hover directions with a paired Legend: this mark's own hover highlighting the
+  // legend (generic mark-hover loop in legendUtils.ts) and a legend hover fading this mark's arcs
+  // (getLegendHighlightOpacityRules in donutUtils.ts) - both match on datum[SERIES_ID], which donut's
+  // rows don't otherwise have, unlike Line/Bar.
```
c-lamoureux: I think this comment is overexplaining.

c-lamoureux (dismissed):

c-lamoureux (approved):

PR #901, opened 2026-09-01
feat(donut-s2): add AdvancedLabel (swatch + name/value/detail block)

madelineluke:
## Summary
- Adds a new `AdvancedLabel` Donut child component - a color swatch + segment name row, optional bold value/percent row, and optional "X out of Y" detail row, positioned around the ring, per `planning/specs/donut/implemented/donut-advanced-labels.json`.
- Extends the direct-label hemisphere-mirrored anchor technique (from `SegmentLabel`) to a 4-part swatch+text block, including manually replicating Vega's polar-to-cartesian math for the swatch since symbol marks have no `radius`/`theta`/`dx` encode channels.
- Reserves ring-gap radius when an `AdvancedLabel` is present (previously only `SegmentLabel` triggered this), and caps the block's vertical row-stacking height using the same existing horizontal-reach ratio, so a segment anchored near the ring's top/bottom never overflows without shrinking the ring further than direct labels already do.
- Generalizes `getSegmentLabelValueFill` into a shared `getLabelValueFill(donutOptions, restColor)` so both direct and advanced labels reuse the same hover-color-switch logic, and extends `isInteractive()` to treat a value/percent-showing `AdvancedLabel` as interactive.
- Fixes a `sanitizeChildren` allowlist bug found during implementation: it maintains its own independent `validDisplayNames` set separate from `childrenAdapter.ts`'s switch, which was silently dropping `AdvancedLabel` children before they ever reached the adapter.
- Collision avoidance between adjacent advanced-label blocks (`donut-direct-labels-collision`) was explicitly deferred and is **not** integrated in this PR - documented as a known limitation in the reconciled spec.

## Test plan
- [x] `yarn test` - 4224 passed (41 new in `advancedLabelUtils.test.tsx`, plus new coverage in `markUtils.test.ts` and `donutSpecBuilder.test.ts`)
- [x] `yarn tsc --noEmit` - no new errors
- [x] `yarn lint` - clean
- [x] Visually verified in Storybook across the `Basic`, `NameOnly`, and a new responsive `WithDetail` story (width-slider, mirroring `DirectLabels.story.tsx`) spanning all size tiers (60px-674px) - confirmed the vertical-overflow fix holds (no label escapes its reserved margin) and the ring stays full-size; XS/S-tier adjacent-label overlap is a known, separately-tracked limitation (no collision avoidance yet)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


madelineluke:
**Follow-up from live testing:** confirmed two real layout issues not caught by unit tests:

- A right-hemisphere label block wider than the reserved horizontal margin overflows the container's edge outright (repro'd at container width 338px / 184px outer diameter, L tier) - there's no horizontal equivalent of the vertical row-stacking cap.
- The vertical row-stacking cap (`getAdvancedLabelRowDy`) does prevent vertical overflow, but as a side effect it compresses the gaps between name/value/detail rows, visibly "smooshing" them together at small sizes when content is oversized.

Tried a horizontal fix mirroring the left hemisphere's capped pull-back - it kept content on-canvas, but did so by shrinking the fixed 20px ring-gap on oversized right-hemisphere labels, which isn't acceptable. Reverted.

Decision: defer both issues to `donut-direct-labels-collision` as a single unit of work, since a correct fix for either likely needs a continuous font-size shrink factor applied to the whole block when it doesn't fit, not a position or spacing adjustment (which is what both attempts here amounted to, each trading away a different invariant). Documented in `planning/specs/donut/implemented/donut-advanced-labels.json`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

packages/vega-spec-builder-s2/src/donut/advancedLabelUtils.ts
```diff
@@ -0,0 +1,488 @@
...
+    },
+  ];
+};
+
+/**
+ * Gets the signals that resolve a donut's advanced-label font sizes from its outer diameter
+ * @param donutOptions
+ * @returns Signal[]
+ */
+export const getAdvancedLabelSignals = (donutOptions: DonutSpecOptions): Signal[] => {
```
c-lamoureux, line 112: Why did we decide to base size scaling on diameter rather than chart width?

c-lamoureux (approved):

PR #912, opened 2026-09-10
fix(s2-donut): create slice gaps with a border

madelineluke:
## Description

To create space between each donut segment, create a white border (background halo) around each slice, rather than using an actual pixel gap. 

## Motivation and Context

Cleaner way to create gaps.

## How Has This Been Tested?

Visually in storybook, unit tests. 

## Types of changes

## Checklist:

### DIFF
--- a/packages/constants/constants.ts
+++ b/packages/constants/constants.ts
@@ -253,14 +253,7 @@ export const DONUT_SIZE_TIER_CUTPOINTS = [120, 160, 200, 400];
 /** S2 donut fixed ring (hole) width per named size tier (XS/S/M/L/XL), used only when holeRatio is left at DEFAULT_HOLE_RATIO */
 export const DONUT_RING_WIDTHS = [16, 18, 20, 22, 24];
 /** S2 donut gap between adjacent segments per named size tier (XS/S/M/L/XL), per chart.donut.size.slice-gap */
-export const DONUT_SLICE_GAPS = [1, 2, 2, 2, 4];
-/**
- * Max fraction of a donut's average segment angular width the slice gap (padAngle) may consume.
- * Without this, a donut with many segments (some much smaller than the average) can have its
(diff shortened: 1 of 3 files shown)


packages/constants/constants.ts
```diff
@@ -253,14 +253,7 @@ export const DONUT_SIZE_TIER_CUTPOINTS = [120, 160, 200, 400];
...
 /** S2 donut gap between adjacent segments per named size tier (XS/S/M/L/XL), per chart.donut.size.slice-gap */
-export const DONUT_SLICE_GAPS = [1, 2, 2, 2, 4];
-/**
- * Max fraction of a donut's average segment angular width the slice gap (padAngle) may consume.
- * Without this, a donut with many segments (some much smaller than the average) can have its
- * smallest segments' entire angular width eaten by the fixed per-tier gap, collapsing them to
- * nothing. The gap degrades gracefully instead by shrinking below the fixed size when needed.
- */
-export const DONUT_SLICE_GAP_MAX_SEGMENT_FRACTION = 1 / 3;
+export const DONUT_SLICE_GAPS = [1, 2, 2, 2, 2];
```
madelineluke, line 256: Right now I have the gap at 2px across the board except for XS. I'm not sure if the gap should ever be 1.5 for some of the sizes?

c-lamoureux: I don't remember us deciding to drop the scaling gap alongside converting it to a border. Am I just forgetting?

madelineluke: originally it was written as [1, 2, 2, 2, 4]. When Alan saw XL he mentioned it should never go above 2. So now it doesn't really scale at all after I dropped XL's gap size from 4 to 2. I'm not sure if I had those values wrong though, and it should be scaling between 1-2 for S-L?

c-lamoureux: Ah yeah. I remember now. Good as is.

c-lamoureux (approved):

